PR #1, opened 2018-08-27
Jona

jonaraphael:
This Pull Request is the first pass at a library of reusable functions for Jupyter.
To see an example notebook look at "2018-08-18 JupyterLib.ipynb"


dubchuck:
General high level comment:  Usage documentation should probably be outlined in the README file.

jason-curtis (changes requested):
I appreciate your desire for version control above what we currently get through Google Drive and your eagerness to get something done. I will bring this up with the software team to potentially fast-track some work on it in lieu of other things we are working on. In the meantime, please follow this ticket and engage with the discussion there to make sure your needs are explained on it:

https://osmobot.atlassian.net/browse/SOFT-465

If we make a library, we need to think through and structure it for reuse. to begin with, putting everything in one file is not an acceptable practice, maintainability-wise.

We also have standards for branch naming, variable naming, test coverage, global variable usage and general cleanliness that will be minimums for any library that we want to maintain and reuse, though the software team hasn't yet made a decision about which we would want to put in place for a jupyter support library. Standards are different in Jupyter-land because of the high level of transparency and iteration that happens there as well as the limitations of the medium, and so that we can have a low barrier to entry for folks like you who want to do experiments but don't have years of professional software under their belt. We don't want a library with those same standards - at least without first having a clear discussion about it. There are also users other than you who helped develop the current process and should have input into any improvements in our experimentation tooling.

again, please engage with the JIRA process that has been working well for the software team. I want to make sure your needs are understood and supported. I mentioned you last week on this ticket which aims to clearly state those needs so that we can make the right moves: https://osmobot.atlassian.net/browse/SOFT-465

jason-curtis:
closing this as osmo-jupyter has superceded this.

PR #2, opened 2018-08-29
SOFT-465 starting to show what a library might look like

jason-curtis:
This and other options discussed in the [design spike doc](https://docs.google.com/document/d/16t_0nefYvwrD542rRH0q5rOfoT9nnql0Ki9na0A5l5E/edit#) . Reading that is not a prerequisite for this PR though it is recommended reading.

There are some particular TODO items in here marked as feedback reqyests from y'all.


osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    source_table = (
+        'calculation_detail join reading on reading.reading_id = calculation_detail.reading_id'
+        if include_hub_id
+        else 'calculation_detail'
+    )
+    nodes_selector = '({})'.format(', '.join(str(n) for n in node_ids)
+
+    calculation_details = pd.read_sql(
+        f'''
```
jason-curtis: looking into fixing my line-ending settings...

README.md
```diff
@@ -1,130 +1,61 @@
...
-- Technician Access to Calculation Details
--- Pulling calculation details from the DB
---- The calculation details table has all kinds of handy internal intermediate details for each calculation.
---- Getting the data from the DB as part of your notebook means you can get updated data just by re-running the notebook. However, since the database table is subject to change, it's also smart to save off the data in a file alongside the notebook so that you can run the notebook later without requiring database access.
---- Recommended practice is to use the database code while the experiment is in progress or when doing a single data pull, but always save that data off to a CSV and switch to using the CSV before you archive the notebook.
-
-# End Jason Notebook
+# Jupyter Notebook Tools
+This is a set of tools that is designed to facilitate science in the context of Osmo Systems.
+The code herein generally expects to be run in the context of a jupyter notebook. TODO (PR feedback plz): eventually it might be used elsewhere. is it OK to put it all here for now and potentially have to refactor/move it later?
```
dubchuck: This seems most appropriate as the location for now as our usage patterns have not emerged outside of the need of Jupyter notebook experimentation.

README.md
```diff
@@ -1,130 +1,61 @@
...
+Function documentation should be in the [Google Style](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings).
+
+### Unit Tests
+
+Run tests with `python setup.py test`
+Any function that has a significant logic component that can be tested, should be unit tested.
+
+# TODO  (PR feedback plz)
+before this library is "ready to go" (or possibly after, if we decide). CR folks, these are questions for you: should we do these now or later
+* fill out database-access functions
```
dubchuck: What do you mean by fill out?

README.md
```diff
@@ -1,130 +1,61 @@
...
+* actually test if I can consume it (might need some kind of GitHub auth setup)
+* implement unit test framework
+* implement linting
```
dubchuck: I vote for starting this now if it's not too involved with regards to what style to use.

README.md
```diff
@@ -1,130 +1,61 @@
...
+### Unit Tests
+
+Run tests with `python setup.py test`
+Any function that has a significant logic component that can be tested, should be unit tested.
+
+# TODO  (PR feedback plz)
+before this library is "ready to go" (or possibly after, if we decide). CR folks, these are questions for you: should we do these now or later
+* fill out database-access functions
+* actually test if I can consume it (might need some kind of GitHub auth setup)
+* implement unit test framework
```
dubchuck: As we are moving towards supporting this as a software team initiative we should include some sort of unit testing.  Perhaps a small subset that is specific to our db access/dataset retrieval will suffice for now?

README.md
```diff
@@ -1,130 +1,61 @@
...
+## Using this library
+
+To consume this library in your notebook, add this line to the top:
+
+`!pip install https://github.com/osmosystems/python-jupyter-tools.git@[VERSION]`
+Where [VERSION] is a version number.
+
+TODO: instructions for installing editable for testing/development purposes. include autoreload magics
+
+## Contributing
```
dubchuck: Do we need a ## Versioning subsection as well?

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
...
+    Raises:
+        sqlalchemy.OperationalError: database connection is not working (often due to wrong password or network disconnect). In this case, a good debugging step is to reconnect to the database.
+    """
+
+    connection = db_engine.connect()
+
+    # TODO? (PR feedback plz): it feels like there's already a fair amount of logic in constructing this query; I could factor out a construct_node_data_query() function here for unit testing
+    downsample_clause = f'AND MOD(calculation_detail.reading_id, {downsample_interval}) = 0' if downsample_interval is not None else ''
+
+    select_clause = (
```
dubchuck: Planting a seed for an ORM of some sort.  Not needed now.

dubchuck: Another solution is to consume the existing API endpoints with curl (or other lib) as opposed to writing db access specifically for this library.

osmo_jupyter/db_access.py
dubchuck: Looks like github diff is barfing on the ```^M``` chars

setup.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+import os
+from distutils.core import setup
+
+setup(
+    name='osmo_jupyter',
```
dubchuck: Name should probably be inline with the repo

README.md
```diff
@@ -1,130 +1,61 @@
...
+Run tests with `python setup.py test`
+Any function that has a significant logic component that can be tested, should be unit tested.
+
+# TODO  (PR feedback plz)
+before this library is "ready to go" (or possibly after, if we decide). CR folks, these are questions for you: should we do these now or later
+* fill out database-access functions
+* actually test if I can consume it (might need some kind of GitHub auth setup)
+* implement unit test framework
+* implement linting
+* fire up CircleCI so that we can see if unit tests and linting are passing on pull requests
```
dubchuck: Leave this as a TODO?  It seems that this could balloon into a larger effort far removed from the initial purpose of this PR.

README.md
```diff
@@ -1,130 +1,61 @@
...
+
+### Unit Tests
+
+Run tests with `python setup.py test`
+Any function that has a significant logic component that can be tested, should be unit tested.
+
+# TODO  (PR feedback plz)
+before this library is "ready to go" (or possibly after, if we decide). CR folks, these are questions for you: should we do these now or later
+* fill out database-access functions
+* actually test if I can consume it (might need some kind of GitHub auth setup)
```
dubchuck: I think pairing this with the CircleCI work below makes sense and possibly can be a followup TODO outside the scope of this PR.

dubchuck (approved):
Looking good so far!  Added some notes/questions.

dubchuck (commented):
Looking good so far!  Added some notes/questions.

README.md
```diff
@@ -1,130 +1,61 @@
...
+
+* "get the standard deviation of a set of node data". Use `my_dataframe['my_column'].std()` as [implemented in pandas](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.std.html) instead.
+* "check temperature and time correlation of each color count for each node during an experiment". This is probably too use-case specific and will end up as unmaintained, "dead code".
+
+## Using this library
+
+To consume this library in your notebook, add this line to the top:
+
+`!pip install https://github.com/osmosystems/python-jupyter-tools.git@[VERSION]`
+Where [VERSION] is a version number.
```
jaimemarijke: s/VERSION/CHANGESET ?

README.md
```diff
@@ -1,130 +1,61 @@
...
+* Code in this library will be reused more, and thus it is more important for it to be correct and even more important for it to be maintainable
+
+### Code Style
+
+This repo follows the Osmo [Code Style Manifesto](https://docs.google.com/document/d/1W1Ipug8IACL4PfZAq5bQKlmfcJGmHGKNH95_FwJcjaI).
+TODO: move the Python section there from proposal to "done"
+
+#### Documentation
+
+Function documentation should be in the [Google Style](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings).
```
jaimemarijke: Should this be called out in Code Style Manifesto instead of here?

README.md
```diff
@@ -1,130 +1,61 @@
...
+Examples of code that should live here:
+
+* Generalized database access functions
+* Basic plots that are useful for a wide variety of experiments
+* Calibration routines
+* Analysis functions
+
+Examples of code that should probably *not* live here:
+
+* "get the standard deviation of a set of node data". Use `my_dataframe['my_column'].std()` as [implemented in pandas](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.std.html) instead.
```
jaimemarijke: Might want to add an explanation that anything that is not pretty Osmo-specific (like this example), likely already exists in a library somewhere, and we should use that library instead.

README.md
jaimemarijke: As discussed, some more TODOs:
 * File tickets for fast-follow-up work
 * Update / restructure Code-style manifesto as necessary to make it obvious what applies to Jupyter notebooks

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
+
+def connect_to_db():
+    # TODO
+
+def load_calculation_details(db_engine, node_ids, start_utc, end_utc, include_hub_id=False, downsample_interval=None):
+    """ Load node data from the calculation_details table, optionally with hub ID included from the readings table
+
+    Args:
```
jaimemarijke: We didn't really talk about this docstring style in-person. 

It does seem generally nicer than other formats I've seen (e.g. @param or :param, which I never really loved), so I guess I'm okay with it even though I suspect it's less common.

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
...
+    Returns:
+        a pandas.DataFrame of data from the node IDs provided.
+    Raises:
+        sqlalchemy.OperationalError: database connection is not working (often due to wrong password or network disconnect). In this case, a good debugging step is to reconnect to the database.
+    """
+
+    connection = db_engine.connect()
+
+    # TODO? (PR feedback plz): it feels like there's already a fair amount of logic in constructing this query; I could factor out a construct_node_data_query() function here for unit testing
+    downsample_clause = f'AND MOD(calculation_detail.reading_id, {downsample_interval}) = 0' if downsample_interval is not None else ''
```
jaimemarijke: I think I agree with the idea to factor out a `construct_node_data_query` function, or at any rate to do something different here. 

I dislike that this function already has a 2x2 based on some somewhat arbitrary parameters - I'm imagining the number of parameters to this function pretty much growing without bound. 

I wonder if it would be better to have the `query` (or some key part of it) passed in as a parameter, and then have other wrapping functions that can construct the appropriate query based on random things like including the hub_id or not?

setup.py
jaimemarijke: I prefer if the repo and the package have the same name (except python and the darn minus sign, so hyphenated in the repo and underscored in the package name). 

I'm not a huge fan of either name, but I'm having a hard time coming up with a better one. I guess between the two I prefer `osmo_jupyter`

If nothing else, I'd prefer if the repo name didn't have `python` in it, since that should be obvious by opening the repo and looking at the code - or in github it just shows up in the ui.

setup.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+import os
+from distutils.core import setup
+
+setup(
+    name='osmo_jupyter',
+    version='0.0.1',
+    author='Osmo Systems',
+    author_email='dev@osmobot.com',
+    description='Tools for accessing and processing OsmoBot data in an experimental context',
```
jaimemarijke: If I could just think of a way to boil this description down into a three-word package name...

`experimental-data-processing-tools`?

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
+
+def connect_to_db():
+    # TODO
+
+def load_calculation_details(db_engine, node_ids, start_utc, end_utc, include_hub_id=False, downsample_interval=None):
+    """ Load node data from the calculation_details table, optionally with hub ID included from the readings table
+
+    Args:
+        db_engine: database engine created using `connect_to_db`
```
jaimemarijke: Is there an argument for why `connect_to_db()` shouldn't just happen within this function? Do we only want to make the connection once and load data multiple different times within a notebook?

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,50 @@
...
+        else 'calculation_detail.*'
+    )
+    source_table = (
+        'calculation_detail join reading on reading.reading_id = calculation_detail.reading_id'
+        if include_hub_id
+        else 'calculation_detail'
+    )
+    nodes_selector = '({})'.format(', '.join(str(n) for n in node_ids)
+
+    calculation_details = pd.read_sql(
```
jaimemarijke: What is `pd`? 

I'm assuming it is pandas, as in `import pandas as pd` which I see in deleted code but not in the current code. Does this code run?

jaimemarijke (approved):
Looks like a good start, though as you pointed out there's more work to be done.

README.md
jason-curtis: I mean, write them as a first set of implemented functionality.

README.md
jason-curtis: OK, as discussed, I'm putting the "line" for landing this just above CircleCI

README.md
jason-curtis: will do, leaving this TODO "above the line"

README.md
jason-curtis: what would you put under that?

osmo_jupyter/db_access.py
jason-curtis: as discussed, solid ideas to revisit in future iterations

osmo_jupyter/db_access.py
jason-curtis: yeah my Windows is showing...

setup.py
jason-curtis: proposal, after some pondering:
* change repo name to: `osmo-jupyter` ("python" and "tools" are redundant, repo names are generally hyphenated)
* change package name to `osmo_jupyter` (from python you want to `from osmo_jupyter import db_access`. unfortunately you can't use hyphen since that's a minus sign in python 🤣 )

remaining unfortunate-ness: the names don't exactly match. but it comes pretty close.

What do you think?

README.md
jason-curtis: yessir. below the line.

osmo_jupyter/db_access.py
jason-curtis: duplicate conversation

README.md
jason-curtis: yes. moved it.

README.md
jason-curtis: 👍

README.md
jason-curtis: 👍 will do in next changeset

README.md
jason-curtis: will be in next changeset

osmo_jupyter/db_access.py
jason-curtis: hmm, sounds like the ORM that Jeremy was suggesting...
someday?

setup.py
jason-curtis: 👍 , resolving this conversation but keeping the one Jeremy opened above until he signs off

osmo_jupyter/db_access.py
jason-curtis: yes, that's what it is, and no, this code does not run. I'll make sure it runs before I land

osmo_jupyter/db_access.py
jason-curtis: yeah, plus I think I'm gonna prompt the user for the db password when we connect.

setup.py
jason-curtis: not `jupyter_python`? if you don't like that, plz engage in the unresolved conversation above

osmo_jupyter/db_access.py
jaimemarijke: hmm, in my head it was much simpler than `ORM` suggests - more like what you suggested here, but taking it in a slightly different direction to limit the number of random parameters going into `load_calculation_details`. something like:

```
query = construct_node_data_query(node_ids= ..., include_hub_id=.., <other random parameters that get shimmed in here>)

data = load_calculation_details(db_engine= ..., query=query)
```

Maybe that's not helping that much since it's just moving the messiness to a different function, but it feels slightly cleaner to me.

jason-curtis: yeah, I see the draw of what you're saying, but I'd like to keep the number of necessary function calls to a minimum here (since the objective of the library is to simplify things for users).

jaimemarijke: yeah I was kind of having that thought too. I guess if you wanted to spend a little more time thinking about how to factor this in a way that helps pave a good path forward for people tweaking it while still maintaining it's ease-of-use, I would be in favor of that. but if not, I suppose I'm okay with it this way too.

README.md
jason-curtis: moved to an "if I can fit it in" status. might give myself 20 minutes to see if it "just works".

README.md
jason-curtis: roger.

README.md
dubchuck: Makes sense. 👍

README.md
dubchuck: I was thinking something akin to a ```CHANGELOG.MD```.

jason-curtis: ah, right. I was gonna leave that out for now - since we're not doing semver, the utility is a bit less and basically you should just use the git commit log.

setup.py
```diff
@@ -1,26 +1,24 @@
...
-        'numpy',
-        'scipy',
-        'pymysql',
-        'petl',
-        'intervaltree',
-        'requests',
-        'plotly==2',
-    ],
-    test_suite='pytest',
-    tests_require=['pytest'],
```
jason-curtis: changes in this file are mainly line endings (sorry! shouldn't happen again) and removing the `test_suite`, `tests_require`, and unused `os` import.

osmo_jupyter/db_access.py
jason-curtis: fixed in latest changeset

setup.py
dubchuck: It's more accurate and 👍.   Works and a readme should be good enough to inform users of a gap in repo name to understanding of it's purpose (see below).

The new repo name doesn't include that how it is a repo mainly supporting external consumers.  I think python tools in this case does provide that more explicit definition.

jason-curtis:
@jaimemarijke @dubchuck this is ready for another round of reviews, particularly on the `db_access` module and tests.

setup.py
dubchuck: 👍

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,109 @@
+import textwrap
+from getpass import getpass
```
dubchuck: Description here (or elsewhere) would be helpful.

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Returns:
+        sqlalchemy Engine object which can be used with other functions in this module.
+    Raises:
+        ValueError: if connection can't be made, usually because password is incorrect
+    """
+    print('Enter database password: (if you don\'t know it, ask someone who does)')
+    db_engine = sqlalchemy.create_engine(
+        'mysql+pymysql://{user}:{password}@{host}/{dbname}'.format(
+            user='technician',
```
dubchuck, line 25: Env vars?

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,109 @@
...
+            In this case, a good debugging step is to reconnect to the database.
+    """
+
+    connection = db_engine.connect()
+
+    calculation_details = pd.read_sql(
+        _get_calculation_details_query(node_ids, start_utc, end_utc, include_hub_id, downsample_interval),
+        db_engine
+    )
+    connection.close()
```
dubchuck: May want to introduce a custom error class or raise an error here for consumers to control what to do.

dubchuck (commented):
Looking good so far.  A few comments...

osmo_jupyter/db_access.py
jason-curtis: hmm, I think this would be end up being annoying for technicians to have to set env variables on every jupyter invocation.
that said, they're not exactly rarin' to input the db password manually, so if you've got a suggestion on how to make env variables painless in everyday life, I'm all ears.

osmo_jupyter/db_access.py
jason-curtis: you mean, a description of what the file does?
    """ Functions to access node data from the Osmo database
    """
sound good?

osmo_jupyter/db_access.py
dubchuck: Agreed that would be annoying.  Nothing comes to mind for a simple fix.

osmo_jupyter/db_access.py
dubchuck: It wasn't immediately clear why the getpass function was being used.

thatneat:
oh, good point. I'm gonna comment where it's used though, not the import.

osmo_jupyter/db_access.py
jason-curtis: gah, I replied to the notification e-mail and the reply didn't get threaded in in the obvious way. I said:
"oh, good point. I'm gonna comment where it's used though, not the import."

jason-curtis:
@jaimemarijke @dubchuck ready for final review, I think!

a note on where I landed on time zones:

Notebooks and analysts will do everything in Osmo Time (this is what folks have been gravitating to and asking for). This library now provides the ability to convert timeseries that came from the database to local time, and to make UTC database calls based on local-time asks from the user.

.circleci/config.yml
```diff
@@ -0,0 +1,34 @@
...
+    docker:
+      - image: circleci/python:3.7
+
+    steps:
+      - checkout
+      
+      - run:
+          name: install package and dependencies
+          command: |
+            python3 -m venv venv
```
jaimemarijke: kinda confusing to call the venv just `venv`

README.md
```diff
@@ -1,130 +1,56 @@
...
-- Technician Access to Calculation Details
--- Pulling calculation details from the DB
---- The calculation details table has all kinds of handy internal intermediate details for each calculation.
---- Getting the data from the DB as part of your notebook means you can get updated data just by re-running the notebook. However, since the database table is subject to change, it's also smart to save off the data in a file alongside the notebook so that you can run the notebook later without requiring database access.
---- Recommended practice is to use the database code while the experiment is in progress or when doing a single data pull, but always save that data off to a CSV and switch to using the CSV before you archive the notebook.
-
-# End Jason Notebook
+# Jupyter Notebook Tools
+This is a set of tools that is designed to facilitate science in the context of Osmo Systems.
+The code herein generally expects to be run in the context of a jupyter notebook. TODO (PR feedback plz): eventually it might be used elsewhere. is it OK to put it all here for now and potentially have to refactor/move it later?
```
jaimemarijke: Remove the TODO?

README.md
```diff
@@ -1,130 +1,56 @@
...
+
+## What lives here
+
+Here's the basic heuristic for what should live in this library vs. in a notebook.
+
+**Is this function reusable across many experiments/activities, both now and for the foreseeable future, and is it fairly Osmo-specific?**
```
jaimemarijke: Might want to change this to a bullet-list to make it more obvious. Otherwise I could see kind of skimming past it. 

e.g.

```
* Is this function reusable across many experiments/activities?
* Will that still be true for the foreseeable future?
* Is it fairly Osmo-specific?
```

README.md
```diff
@@ -1,130 +1,56 @@
...
+
+### Code Style
+
+This repo follows the Osmo [Code Style Manifesto](https://docs.google.com/document/d/1W1Ipug8IACL4PfZAq5bQKlmfcJGmHGKNH95_FwJcjaI).
+
+
+### Installing for development
+
+1. Check out the repository and switch to its directory
+2. `pip install --editable .`
```
jaimemarijke: Might want to explain what this is, otherwise it almost reads as what you type to accomplish #1.

.circleci/config.yml
```diff
@@ -0,0 +1,34 @@
...
+            . venv/bin/activate
+            pip install .
+            pip install pytest flake8
```
jaimemarijke: can't these be installed as test dependencies somehow?

.circleci/config.yml
jaimemarijke: also, I'm not sure I understand why we need a venv at all. isn't circleci kind of a big venv?

osmo_jupyter/timezone.py
```diff
@@ -0,0 +1,51 @@
...
+    """ get a timezone-aware object from local time string(s).
+
+    Args:
+        local_time: string of local time in any valid non-timezone-aware ISO format
+            Time should be in Osmo HQ local time.
+            eg. ('2012-01-01 12:00:00', '2012-01-01 12:00', '2012-01-01')
+    Returns:
+        timezone-aware datetime object
+    """
+    time = datetime.datetime.fromisoformat(local_time)
```
jason-curtis: hmmmm. `fromisoformat()` turns out to be a python 3.7-only thing. most folks have been using python 3.6. currently it is possible to install python 3.7 in anaconda, but there isn't a [distribution](https://www.anaconda.com/download/#linux) that comes with it preinstalled. 
input needed: should I:
1. do this using dateutil instead, which will allow a wider range of dates (too permissive, really - it'll make a "guess" on some crazy stuff) or
2. take a dependency on [iso8601](https://pypi.org/project/iso8601/) and use that, or
3. just "hold the line" on python 3.7 and teach people how to upgrade/hope a distro comes out soon

.circleci/config.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - run:
+          name: tests
+          command: |
+            . venv/bin/activate
+            pytest
```
jaimemarijke: could use tox to run both pytest and flake

README.md
```diff
@@ -1,130 +1,56 @@
...
+* "get the standard deviation of a set of node data". Use `my_dataframe['my_column'].std()` as [implemented in pandas](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.std.html) instead. Anything that is not Osmo-specific is probably already implemented in an existing library that we should use instead.
+* "check temperature and time correlation of each color count for each node during an experiment". This is probably too use-case specific and will end up as unmaintained, "dead code".
+
+## Using this library
+
+Before you can consume the library, you'll need to [set up an SSH key with GitHub](https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/). This lets your computer access the private Git repository and you may have already done this.
+
+Then, add a line near the top of your notebook:
+`!pip install https://github.com/osmosystems/osmo-jupyter.git@[CHANGESET]`
+Where [CHANGESET] is a changeset ID. Grab the latest changeset ID by [visiting GitHub](https://github.com/OsmoSystems/osmo-jupyter) and searching for "latest commit". Copy that value.
```
jaimemarijke, line 32: Could add an explanation of why this is important

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,128 @@
...
+import sqlalchemy
+
+import textwrap
+from getpass import getpass
+from osmo_jupyter import timezone
+
+
+def configure_database():
+    """ Configure a database object for read-only technician access to Osmo data.
+    Requries user input to collect database password.
```
jaimemarijke: typo: Requires

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,128 @@
...
+        start_time_local: string of ISO-formatted start datetime in local time, inclusive
+        end_time_local: string of ISO-formatted end datetime in local time, inclusive
+        include_hub_id: if True, the output will include a 'hub_id' column.
+            Default False because the request including hub_id takes extra time.
+        downsample_factor: if this is a number, it will be used to select fewer rows.
+            You should get *roughly* n / downsample_factor samples.
+    Returns:
+        a pandas.DataFrame of data from the node IDs provided.
+    Raises:
+        sqlalchemy.OperationalError: database connection is not working
```
jaimemarijke, line 150: Looks like this will actually return a ValueError since you catch the OperationalError

osmo_jupyter/db_access.py
```diff
@@ -0,0 +1,128 @@
...
+            This is often due to wrong password or network disconnect.
+            In this case, a good debugging step is to reconnect to the database.
+    """
+
+    connection = db_engine.connect()
+
+    calculation_details = pd.read_sql(
```
jaimemarijke, line 157: side-note: it is somehow surprising to me that pandas knows how to talk to a sql database. But I guess pandas is just good at data, and part of that is good at talking to dbs?

osmo_jupyter/db_access_test.py
```diff
@@ -0,0 +1,50 @@
...
+start_utc = '2018-08-08 19:00:00'
+end_utc = '2018-09-09 20:00:00'
+
+
+@pytest.mark.parametrize(
+    'name, include_hub_id, downsample_factor, query_contents_to_look_for, should_be_present',
+    [
+        ('base case', False, None, 'SELECT', True),
+        ('SELECT clause with both options on', True, 10, 'SELECT calculation_detail.*, reading.hub_id', True),
+        ('MOD clause for downsample option', True, 10, 'AND MOD(calculation_detail.reading_id, 10) = 0', True),
```
jaimemarijke: typo: MOD clause *present* for ...

osmo_jupyter/db_access_test.py
```diff
@@ -0,0 +1,50 @@
...
+import textwrap
+
+
+node_ids = [123, 456]
+start_utc = '2018-08-08 19:00:00'
+end_utc = '2018-09-09 20:00:00'
+
+
+@pytest.mark.parametrize(
+    'name, include_hub_id, downsample_factor, query_contents_to_look_for, should_be_present',
```
jaimemarijke, line 13: feels like kind of a weird test to be selectively looking for SQL statement fragments, but I guess that's pretty much what you've got to work with.

osmo_jupyter/timezone.py
```diff
@@ -0,0 +1,51 @@
...
+
+import pytz
+import pandas as pd
+
+
+OSMO_HQ_TIMEZONE = pytz.timezone('US/Pacific')
+SQL_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'  # Time format favored by mysql
+
+
+def to_aware(local_time):
```
jaimemarijke: should this have a usage example?

jaimemarijke: same for the other functions here

osmo_jupyter/timezone.py
```diff
@@ -0,0 +1,51 @@
+import datetime
+
+import pytz
+import pandas as pd
+
+
+OSMO_HQ_TIMEZONE = pytz.timezone('US/Pacific')
```
jaimemarijke, line 5: does this cover DST too?

osmo_jupyter/timezone.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def utc_series_to_local(pandas_timeseries):
+    ''' Convert a Pandas series of UTC times to local time
+    Useful for converting node data timestamps for use in local-time notebooks.
+
+    Args:
+        pandas_timeseries: pandas Series of timezone-naive datetimes or strings in UTC
+    Returns:
+        Pandas timeseries of timezone-naive datetimes corresponding to Osmo HQ local time
```
jaimemarijke, line 19: why leave them timezone naive? why not set the timezone to Osmo HQ?

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[flake8]
+ignore = D203
```
jaimemarijke: comment?

osmo_jupyter/timezone_test.py
```diff
@@ -0,0 +1,37 @@
...
+import pytest
+import pandas as pd
+
+from osmo_jupyter import timezone as module
+
+
+@pytest.mark.parametrize('input_datetime, expected_iso_output', [
+    ('2018-01-01 12:00', '2018-01-01T12:00:00-08:00'),
+    ('2018-06-01 12:00', '2018-06-01T12:00:00-07:00'),
+])
```
jaimemarijke: nice! parameterized test for DST ftw

osmo_jupyter/timezone_test.py
```diff
@@ -0,0 +1,37 @@
...
+def test_to_aware(input_datetime, expected_iso_output):
+    output_datetime = module.to_aware(input_datetime)
+
+    assert isinstance(output_datetime, datetime.datetime)
+    assert output_datetime.isoformat() == expected_iso_output
+
+
+def test_to_aware_blows_up_if_timezone_provided():
```
jaimemarijke: kind of weird to test the functionality of the 3rd party library

jaimemarijke: but I guess if it's functionality we want, maybe it's nice to call that out?

osmo_jupyter/timezone_test.py
```diff
@@ -0,0 +1,37 @@
...
+    time_string = '2018-01-01 12:00Z'
+    with pytest.raises(ValueError):
+        module.to_aware(time_string)
+
+
+def test_to_utc_string():
+    time_string = '2018-01-01 01:11'
+    assert module.to_utc_string(time_string) == '2018-01-01 09:11:00'
```
jaimemarijke: should this also have parameterized tests from different DST dates?

osmo_jupyter/timezone.py
```diff
@@ -0,0 +1,51 @@
...
+            Time should be in Osmo HQ local time.
+            eg. ('2012-01-01 12:00:00', '2012-01-01 12:00', '2012-01-01')
+    Returns:
+        timezone-aware datetime object
+    """
+    time = datetime.datetime.fromisoformat(local_time)
+    return OSMO_HQ_TIMEZONE.localize(time)
+
+
+def to_utc_string(local_time):
```
jaimemarijke: I wonder if all of these function names should have `osmo` in them, or something, to call out that they're hardcoded to Osmo HQ timezone.

That or maybe take the timezone as a param with Osmo as the default (though maybe that's just dangerous?)

Huh, wonder if Jona's going to be running any experiments from Boston?

osmo_jupyter/timezone_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_utc_series_to_local():
+    utc_series = pd.Series(['2018-01-01 10:00', '2018-07-01 20:00'])
+    expected_local_time_series = pd.to_datetime(pd.Series(['2018-01-01 18:00:00', '2018-07-02 03:00:00']))
```
jaimemarijke: ehh, kind of weird for the test to actually replicate the call in the function? can this be hardcoded instead?

osmo_jupyter/timezone_test.py
```diff
@@ -0,0 +1,37 @@
...
+    
+    actual_local_time_series = module.utc_series_to_local(utc_series)
+
+    pd.testing.assert_series_equal(actual_local_time_series, expected_local_time_series)
```
jaimemarijke: pd.testing!??

osmo_jupyter/timezone.py
jaimemarijke: especially since, as a non-dev, `to_aware` might not mean all that much as a function name.

Though I guess maybe the main usage of functions from this module will be `utc_series_to_local`?

jaimemarijke (approved):
Mostly nitpicky stuff.

Ahhhh python.

.circleci/config.yml
jason-curtis: yeah, I didn't have much luck with putting them in setup.py. distutils is kind of a mess.
moved to `test_requirements.txt`

osmo_jupyter/timezone_test.py
jason-curtis: yeah I was going to write the test anyway before I realized it was already implemented. should prevent us accidentally removing the functionality

osmo_jupyter/timezone_test.py
jason-curtis: could be, but it's depending on the other function. didn't really want to recreate that whole test.

osmo_jupyter/timezone.py
jason-curtis: I'll follow up with Jona. i think this is the right level of specificity for now, we can add kwargs later if necessary. actually started with kwargs but backtracked because YAGNI

osmo_jupyter/timezone_test.py
jason-curtis: replaced with direct `datetime.datetime` calls

osmo_jupyter/timezone_test.py
jason-curtis: 💥 
yeah, otherwise you get issues where "the truth value of a Series is undefined. please use Series.all() or [blah blah blah]".
`pd.testing` does reasonable things.

.circleci/config.yml
jason-curtis: I think otherwise you have to sudo pip install? this was copied from an example on circleci's site. I'll check after it's in master - circleci doesn't seem to be picking it up from the branch.

osmo_jupyter/timezone.py
dubchuck: 3) seems like it could drop down a rabbit hole in helping with individual configurations for 8 people and the other issues that might arise with using python 3.7 within anaconda.

For 1) Can you give an example of what it would guess on?

For now I'd vote for option 2 given what seems like reliable functionality (cursory glance) and doesn't probably introduces less "unknown unknowns".

.circleci/config.yml
dubchuck: You could also switch to our a docker container registry and use a pre configured image.

.circleci/config.yml
```diff
@@ -16,19 +16,19 @@ jobs:
       - run:
           name: install package and dependencies
           command: |
-            python3 -m venv venv
-            . venv/bin/activate
+            python3 -m venv my_venv
+            . my_venv/bin/activate
```
jaimemarijke: maybe `test_venv`?

README.md
```diff
@@ -1,13 +1,15 @@
...
 ## What lives here
 
 Here's the basic heuristic for what should live in this library vs. in a notebook.
 
-**Is this function reusable across many experiments/activities, both now and for the foreseeable future, and is it fairly Osmo-specific?**
-yes = probably it should live here. no = probably not.
+* Is this function reusable across many experiments/activities?
+* Will that still be true for the foreseeable future?
+* Is it fairly Osmo-specific?
+**yes** = probably it should live here. **no** = probably not.
```
jaimemarijke: **yes to all**
**no to any**

osmo_jupyter/timezone.py
jaimemarijke: huh is `iso8601` new? I never loved `dateutil`, so I guess I'd be up for something else

so, seconding the vote for #2

osmo_jupyter/timezone.py
dubchuck: +1 to function name and usage ex

setup.cfg
jaimemarijke: apparently this rule is an error code from [pydocstyle](https://pep257.readthedocs.io/en/latest/index.html)'s implementation of [PEP-257](https://www.python.org/dev/peps/pep-0257/).

D203 "1 blank line required before class docstring" 
However, apparently it's also in direct conflict with D211 "No blank lines allowed before class docstring", and supposedly is disabled by default. So maybe we can just drop this ignore?

I just looked at our standard flake8 ignores from my team at ES, which seem to come from a mix of different rule sets:

E501 - line too long (>79 char) ([pep8](https://pep8.readthedocs.io/en/latest/intro.html#error-codes))
F403 - wildcard imports ([pyflakes](http://flake8.pycqa.org/en/2.5.5/warnings.html))
F999 - ???
S101 - Multi-line construct missing trailing comma ([flake8_strict](https://pypi.org/project/flake8_strict/))

Of those, the only one that really seems relevant is `E501`

jaimemarijke (approved):

osmo_jupyter/db_access.py
jason-curtis: good point, also updating the docstring below to reflect that we already checked the password

osmo_jupyter/db_access.py
jason-curtis: yep. it speaks Excel and HDFS too.

osmo_jupyter/db_access_test.py
jason-curtis: comin *riight* up against that 120 limit ;)

osmo_jupyter/db_access_test.py
jason-curtis: yep

osmo_jupyter/timezone.py
jason-curtis: yeah, this is kinda weird because the only function in here I really expect folks to be using is `utc_series_to_local`. I'm finally convinced - gonna move those other two functions into `db_access` and make them private.

osmo_jupyter/timezone.py
jason-curtis: yes, looks like you saw this below

osmo_jupyter/timezone.py
jason-curtis: because Plotly always plots datetimes as if they were in UTC :/
ref: https://community.plot.ly/t/timezone-on-plotlyjs/26

setup.cfg
jason-curtis: yeah and that one's covered by setting `max_line_length = 120`

.circleci/config.yml
jason-curtis: I'd like the requirements to be installed according to the files in the repo, so I don't think there's anything that we'd benefit from prebuilding here. we could potentially cache some eggs or something though, using the caching feature.

PR #4, opened 2018-08-31
time zones were backwards!

jason-curtis:
I goofed.

### DIFF
--- a/osmo_jupyter/db_access.py
+++ b/osmo_jupyter/db_access.py
@@ -95,8 +95,8 @@ def _get_calculation_details_query(
         SQL query that can be used to get the desired data
     """
 
-    start_timezone_aware_string = _to_utc_string(start_time_local)
-    end_timezone_aware_string = _to_utc_string(end_time_local)
+    start_utc_string = _to_utc_string(start_time_local)
+    end_utc_string = _to_utc_string(end_time_local)
     downsample_clause = (
         f'AND MOD(calculation_detail.reading_id, {downsample_factor}) = 0'
         if downsample_factor is not None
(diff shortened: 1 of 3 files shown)


osmo_jupyter/db_access.py
```diff
@@ -95,8 +95,8 @@ def _get_calculation_details_query(
         SQL query that can be used to get the desired data
     """
 
-    start_timezone_aware_string = _to_utc_string(start_time_local)
-    end_timezone_aware_string = _to_utc_string(end_time_local)
+    start_utc_string = _to_utc_string(start_time_local)
```
jason-curtis, line 98: stuff in this file wasn't really a bug, just incorrect variable names

osmo_jupyter/timezone.py
```diff
@@ -20,5 +20,5 @@ def utc_series_to_local(pandas_timeseries):
     '''
     definitely_a_timeseries = pd.to_datetime(pandas_timeseries)
 
-    localized_timeseries = definitely_a_timeseries.dt.tz_localize(OSMO_HQ_TIMEZONE)
-    return localized_timeseries.dt.tz_convert('UTC').dt.tz_localize(None)
+    localized_timeseries = definitely_a_timeseries.dt.tz_localize('UTC')
```
jason-curtis, line 24: this was previously converting local to UTC, not UTC to local :p

osmo_jupyter/timezone.py
```diff
@@ -20,5 +20,5 @@ def utc_series_to_local(pandas_timeseries):
     '''
     definitely_a_timeseries = pd.to_datetime(pandas_timeseries)
```
dubchuck: Fun but maybe something more definite...err...why is it definitely_a_timeseries?

dubchuck (approved):
One question, otherwise LGTM.

osmo_jupyter/timezone.py
jason-curtis: because the input may have been a series of strings. I had been meaning to clarify that.

PR #5, opened 2018-09-26
Initial set of plotting tools

jason-curtis:
Check out the example notebook that shows how this stuff works, under jupyter snippets/osmo-jupyter plotting and normalization

### DIFF
--- a/osmo_jupyter/normalize_test.py
+++ b/osmo_jupyter/normalize_test.py
@@ -14,7 +14,7 @@ def test_normalize_to_max():
     })
 
     actual = module.to_max(dataframe)
-    assert pd.testing.assert_frame_equal(actual, expected)
+    pd.testing.assert_frame_equal(actual, expected)
 
 
 def test_normalize_by_rgb_sum():
@@ -29,5 +29,5 @@ def test_normalize_by_rgb_sum():
         'b': [8/14, 10/18, 9/19],
     })
 
-    actual = module.to_max(dataframe)
-    assert pd.testing.assert_frame_equal(actual, expected)
+    actual = module.by_rgb_sum(dataframe)
+    pd.testing.assert_frame_equal(actual, expected)
--- a/osmo_jupyter/normalize_test.py
+++ b/osmo_jupyter/normalize_test.py
@@ -1,0 +1,33 @@
+import pandas as pd
+
+import .normalize as module
+
+
+def test_normalize_to_max():
+    dataframe = pd.DataFrame({
+        'column a': [1, 2, 3],
+        'column b': [8, 10, 9],
+    })
+    expected = pd.DataFrame({
+        'column a': [1/3, 2/3, 1],
+        'column b': [.8, 1, .9],
+    })
+
+    actual = module.to_max(dataframe)
+    assert pd.testing.assert_frame_equal(actual, expected)
+
+
+def test_normalize_by_rgb_sum():
+    dataframe = pd.DataFrame({
+        'r': [1, 2, 3],
+        'g': [5, 6, 7],
+        'b': [8, 10, 9],
+    })
+    expected = pd.DataFrame({
+        'r': [1/14, 2/18, 3/19],
+        'g': [5/14, 6/18, 7/19],
+        'b': [8/14, 10/18, 9/19],
+    })
+
+    actual = module.to_max(dataframe)
+    assert pd.testing.assert_frame_equal(actual, expected)
--- a/osmo_jupyter/normalize_test.py
+++ b/osmo_jupyter/normalize_test.py
@@ -1,6 +1,6 @@
 import pandas as pd
 
-import .normalize as module
+import osmo_jupyter.normalize as module
 
 
 def test_normalize_to_max():
--- a/osmo_jupyter/plot.py
+++ b/osmo_jupyter/plot.py
@@ -114,7 +114,6 @@ def get_rgb_scatters(
                 **marker_overrides
             },
             opacity=0.8,
-            # yaxis='y3',
             yaxis=f'y{AXES_BY_COLOR[color]}' if colors_on_separate_axes else f'y{SHARED_COLOR_AXIS}'
         )
         for color in colors_to_include
--- a/osmo_jupyter/plot.py
+++ b/osmo_jupyter/plot.py
@@ -176,7 +176,7 @@ def get_rgb_columns(df, pattern, other_columns_to_include=None):
 
 def get_layout_with_annotations(
     x_axis_title='Time',
-    y_axis_title=None,
+    y_axis_title='',
     colors_on_separate_axes=False,
     events=None,
     **additional_layout_kwargs
@@ -199,9 +199,6 @@ def get_layout_with_annotations(
             Valid options are documented at https://plot.ly/python/reference/#layout
     '''
     events = events or {}
-
-    if not colors_on_separate_axes and not y_axis_title:
-        raise ValueError('If colors_on_separate_axes is False you must provide a y_axis_title')
 
     def get_color_y_axis_title(color):
         if not colors_on_separate_axes:
--- a/osmo_jupyter/plot.py
+++ b/osmo_jupyter/plot.py
@@ -1,0 +1,224 @@
+import plotly.graph_objs as go
+
+
+COLOR_CHANNELS = ['r', 'g', 'b']
+
+# Axis # to use for each color when colors are being plotted on separate axes
+AXES_BY_COLOR = {
+    'r': 2,
+    'g': 3,
+    'b': 4,
+}
+ANNOTATION_AXIS = 1
+NON_ANNOTATION_AXIS = 2
+
+
+COLORS_BY_LETTER = {
+    'r': 'red',
+    'g': 'green',
+    'b': 'blue'
+}
+
+
+def subsample_for_plot(df, dots_to_plot=150):
+    ''' resample every nth row so that Plotting doesnt use so much CPU
+    '''
+    n = int(len(df) / dots_to_plot) or 1
+    return df.iloc[::n, :]
+
+
+def get_rgb_scatters(
+    rgb_df,
+    x_column=None,
+    colors_on_separate_axes=False,
+    dataset_name=None,
+    marker_overrides=None,
+    colors_to_include=COLOR_CHANNELS,
+):
+    ''' get go.Scatter instances for each 'rgb' color in the input DataFrame
+
+    example usage:
+    >>> import pandas as  pd
+    >>> import plotly.graph_objs as go
+    >>> rgb_df = pd.DataFrame({'r': [1,2,3], 'g': [2,3,4], 'b': [3,4,5], 'x_axis': [0,1,2]})
+    >>> go.FigureWidget(rgb_scatters(rgb_df, x_column='x_axis'))
+    [result is a plot with your three channels]
+
+    Params:
+        rgb_df: pandas.DataFrame instance with 'r', 'g', and 'b' columns, and a column for the X axis value
+        x_column: Optional, name of the column in rgb_df to use for the X axis values. if None, we'll use the index
+        colors_on_separate_axes: Optional, if True, will plot each color series on a separate axis
+        dataset_name: Optional, string to use in legend labels
+        marker_overrides: Optional, dict of additional arguments
+        colors_to_include: Optional list of color letters to include - provide if you don't want all the colors.
+            Should be a subset of COLOR_CHANNELS.
+    Returns:
+        list of go.Scatter objects, one each for r, g, and b or colors_to_include if provided
+    '''
+    marker_overrides = marker_overrides or {}
+    x_column_values = rgb_df[x_column] if x_column else rgb_df.index
+    return [
+        go.Scatter(
+            x=x_column_values,
+            y=list(rgb_df[color]),
+            name=f'{dataset_name} - {color}' if dataset_name is not None else color,
+            mode='markers',
+            marker={
+                'color': COLORS_BY_LETTER[color],
+                'symbol': 'circle-open',
+                'size': 5,
+                **marker_overrides
+            },
+            opacity=0.8,
+            # yaxis='y3',
+            yaxis=f'y{AXES_BY_COLOR[color]}' if colors_on_separate_axes else f'y{NON_ANNOTATION_AXIS}'
+        )
+        for color in colors_to_include
+    ]
+
+
+def axis_kwargs(axis_number, title, color=None, **y_axis_params):
+    color_kwargs = {
+        'titlefont': {'color': color},
+        'tickfont': {'color': color},
+        'tickcolor': color,
+    } if color else {}
+    return {
+        f'yaxis{axis_number}': {
+            'title': title,
+            **color_kwargs,
+            **y_axis_params
+        }
+    }
+
+
+def _add_dicts(dicts):
+    output_dict = {}
+    for _dict in dicts:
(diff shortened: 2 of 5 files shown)


osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+            name=f'{dataset_name} - {color}' if dataset_name is not None else color,
+            mode='markers',
+            marker={
+                'color': COLORS_BY_LETTER[color],
+                'symbol': 'circle-open',
+                'size': 5,
+                **marker_overrides
+            },
+            opacity=0.8,
+            # yaxis='y3',
```
jaimemarijke: remove?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+    ''' get go.Scatter instances for each 'rgb' color in the input DataFrame
+
+    example usage:
+    >>> import pandas as  pd
+    >>> import plotly.graph_objs as go
+    >>> rgb_df = pd.DataFrame({'r': [1,2,3], 'g': [2,3,4], 'b': [3,4,5], 'x_axis': [0,1,2]})
+    >>> go.FigureWidget(rgb_scatters(rgb_df, x_column='x_axis'))
+    [result is a plot with your three channels]
+
+    Params:
```
jaimemarijke: Should be "Args", right?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+            },
+            opacity=0.8,
+            # yaxis='y3',
+            yaxis=f'y{AXES_BY_COLOR[color]}' if colors_on_separate_axes else f'y{NON_ANNOTATION_AXIS}'
+        )
+        for color in colors_to_include
+    ]
+
+
+def axis_kwargs(axis_number, title, color=None, **y_axis_params):
```
jaimemarijke: should this be private? It's not really clear what it would be used for

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+    return {
+        f'yaxis{axis_number}': {
+            'title': title,
+            **color_kwargs,
+            **y_axis_params
+        }
+    }
+
+
+def _add_dicts(dicts):
```
jaimemarijke, line 142: :/

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+        x_column: Optional, name of the column in rgb_df to use for the X axis values. if None, we'll use the index
+        colors_on_separate_axes: Optional, if True, will plot each color series on a separate axis
+        dataset_name: Optional, string to use in legend labels
+        marker_overrides: Optional, dict of additional arguments
+        colors_to_include: Optional list of color letters to include - provide if you don't want all the colors.
+            Should be a subset of COLOR_CHANNELS.
+    Returns:
+        list of go.Scatter objects, one each for r, g, and b or colors_to_include if provided
+    '''
+    marker_overrides = marker_overrides or {}
```
jaimemarijke, line 108: I remember there's some reason for doing this instead of setting a default parameter value... but I don't remember being that convinced by it at the time.

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+        y_axis_title: title to use for Y axis where colors are plotted. Required if colors_on_separate_axes is False.
+            If colors_on_separate_axes is True, it will be used as a suffix for the y axis titles "r", "g", and "b"
+        colors_on_separate_axes: whether colors (r,g and b) should be displayed on separate Y axes
+        events: dictionary of {x_axis_value: "annotation"} which will be used to annotate the chart
+        additional_layout_params: These will be passed to go.Layout. These can, for instance, be the result of
+            axis_kwargs() to add support for graphing more items.
+    '''
+    events = events or {}
+    if not colors_on_separate_axes and not y_axis_title:
+        raise ValueError('If colors_on_separate_axes is False you must provide a y_axis_title')
```
jaimemarijke: Why is it required?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+            'range': [0, 1],
+            'visible': False,
+        }
+    }
+
+    def get_color_y_axis_title(color):
+        if not colors_on_separate_axes:
+            return y_axis_title
+
+        # "blue" axis uses a line break to not overlap unnecessarily with green
```
jaimemarijke: If I hadn't seen the graph and noticed that red was on the left but both blue and green were on the right, I think I would have been pretty confused by this..

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+            'showarrow': True,
+            'textangle': -55,
+            'ax': 1
+        }
+        for title, x_value in events.items()
+    ]
+
+    return go.Layout(
+        xaxis={'title': 'Time'},
+        annotations=dummy_annotation + event_annotations,
```
jaimemarijke: you could do a dummy annotation or an event_annotation - not sure if it matters

```annotations = event_annotation if event_annotation else dummy_annotation```

File: osmo_jupyter/plot.py [hunk 9]
```diff
@@ -0,0 +1,224 @@
...
+    * The primary Y axis (y0) is reserved for annotations. This is done to guarantee that annotations show
+    * If colors are on separate axes, they will be on the next consecutive axes, otherwise they will be on yaxis 2
+    * You can use additional axes for plotting other things by passing in additional params
+
+    Params:
+        x_axis_title: title to use for X axis, default 'Time'
+        y_axis_title: title to use for Y axis where colors are plotted. Required if colors_on_separate_axes is False.
+            If colors_on_separate_axes is True, it will be used as a suffix for the y axis titles "r", "g", and "b"
+        colors_on_separate_axes: whether colors (r,g and b) should be displayed on separate Y axes
+        events: dictionary of {x_axis_value: "annotation"} which will be used to annotate the chart
```
jaimemarijke: Is this dictionary backwards? isn't "annotation" the key?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,224 @@
...
+            return y_axis_title
+
+        # "blue" axis uses a line break to not overlap unnecessarily with green
+        line_break = '<br>' if color == 'b' else ''
+
+        # If colors are on separate axes and a Y axis title is provided, use it as a suffix
+        suffix = f' {y_axis_title}' if y_axis_title else ''
+        return f'{line_break}{color}{suffix}'
+
+    color_y_axes = _add_dicts(
```
jaimemarijke: this block is a bit hard to parse ... could help to break it out into separate variables, e.g.

```
colors_on_same_axis_kwargs = ...
colors_on_separate_axes_kwargs = ...

color_y_axes = colors_on_separate_axes_kwargs if colors_on_separate_axes else colors_on_same_axis_kwargs
```

jaimemarijke (commented):
partial review - just getting it out for now

osmo_jupyter/plot.py
jason-curtis: :/ and we're so close now with being able to unpack dicts inside other dicts!

osmo_jupyter/plot.py
jason-curtis: fixed

osmo_jupyter/plot.py
jason-curtis: yeah it's not the greatest reason. the problem is that the default value only gets interpreted once, so  *if* the value gets mutated inside **or after** the function call (like say, I passed it out and someone added a key to the dictionary) the mutated value gets used forevermore as the default value for the function.
so this is a best practice... I could be convinced we should just ignore the edge case until we (as this team) have felt the pain :p

osmo_jupyter/plot.py
jason-curtis: because you should always label your axes.
good point though I don't really need to enforce that here; I can just let y_axis title be an empty string by default.

osmo_jupyter/plot.py
jason-curtis: yeah the whole colors on separate axis behavior is kinda gross. I'll see if I can clean it up a bit.

osmo_jupyter/plot.py
jason-curtis: yeah the extra annotation doesn't hurt us so I just left it.

jason-curtis: resisted the temptation to use `annotations=event_annotations or dummy_annotation,` using your suggestion instead

osmo_jupyter/plot.py
jason-curtis: thanks, fixing

osmo_jupyter/plot.py
jason-curtis: doing that. will be in next push

jason-curtis:
Marking this as no longer WIP. There's a usage example in the jupyter snippets folder.
I'll file a separate ticket for plotting values in color space and ask for prioritization from users.

osmo_jupyter/plot.py
jason-curtis: I cleaned it up a little.

osmo_jupyter/normalize_test.py
```diff
@@ -29,5 +29,5 @@ def test_normalize_by_rgb_sum():
         'b': [8/14, 10/18, 9/19],
     })
 
-    actual = module.to_max(dataframe)
-    assert pd.testing.assert_frame_equal(actual, expected)
+    actual = module.by_rgb_sum(dataframe)
```
jaimemarijke, line 32: woah, was this test passing before?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,277 @@
...
+            we'll use the index from it
+        dataset_name: Optional, name to put in the chart legend. If not provided, y_values is expected to be a Series
+            and we'll use its name
+        y_axis: Optional, Y-axis number. Defaults to the primary (1) Y axis. Note that axes 2-4 are used for RGB values.
+        marker_overrides: Optional, dict of go.Scatter.Marker attributes to override marker style.
+            Valid options are documented at https://plot.ly/python/reference/#scatter-marker
+        scatter_overrides: Optional, dict of go.Scatter attributes to override the Scatter.
+            Valid options are documented at https://plot.ly/python/reference/#scatter
+    '''
+    x_column_values = x_values or y_values.index
```
jaimemarijke: Doesn't y_values have to be a dataframe for this to work? Description says "pandas Series or whatever"

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,277 @@
...
+):
+    ''' Quick way to get a go.Scatter to put on your plot with some conveniences.
+
+    Args:
+        y_values: iterable of Y values can be a pandas Series or whatever
+        x_values: Optional, X values to go along with those. If not provided, y_values is expected to be a Series and
+            we'll use the index from it
+        dataset_name: Optional, name to put in the chart legend. If not provided, y_values is expected to be a Series
+            and we'll use its name
+        y_axis: Optional, Y-axis number. Defaults to the primary (1) Y axis. Note that axes 2-4 are used for RGB values.
```
jaimemarijke: maybe call this `y_axis_number`?

osmo_jupyter/plot.py
jaimemarijke: bleh that's right... 

probably best to just leave it as you have it, though I'm also tempted to suggest pyrsistent..

osmo_jupyter/plot.py
jaimemarijke: yeah seems kind of annoying to enforce that when you might just want to throw a quick graph together

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,277 @@
...
+        'yaxis': {
+            'title': '',
+            'visible': False,
+        }
+    }
+    # We need to have at least 1 annotation on axis 1 or else plotted values will disappear
+    # https://github.com/plotly/plotly.py/issues/1200
+    dummy_annotation = [{
+        'yref': 'y',
+        'xref': 'paper',
```
jaimemarijke, line 248: what is 'paper'?

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,277 @@
...
+        'text': '',
+        'showarrow': False,
+        'visible': True,
```
jaimemarijke, line 251: why does it have to be visible?

setup.py
```diff
@@ -14,7 +14,7 @@
         'numpy',
         'pandas',
         'petl',
-        'plotly >= 2, < 3',
+        'plotly >= 3, < 4',
```
jaimemarijke, line 17: woah, taking the leap! Do we need to make sure everyone's on the same page about this? presumably people currently have plotly 2.x installed and expect to use that in their jupyter notebooks.

osmo_jupyter/plot.py
```diff
@@ -0,0 +1,277 @@
...
+            Valid options are documented at https://plot.ly/python/reference/#scatter-marker
+        colors_to_include: Optional list of color letters to include - provide if you don't want all the colors.
+            Should be a subset of COLOR_CHANNELS.
+    Returns:
+        list of go.Scatter objects, one each for r, g, and b or colors_to_include if provided
+    '''
+    marker_overrides = marker_overrides or {}
+    x_column_values = rgb_df[x_column] if x_column else rgb_df.index
+    return [
+        go.Scatter(
```
jaimemarijke: does it make sense to use your other function, `get_scatter` here instead?

jaimemarijke (approved):

osmo_jupyter/normalize_test.py
jason-curtis: nope.

osmo_jupyter/plot.py
jason-curtis: Series also have indexes. doc says `x_values: [...] If not provided, y_values is expected to be a Series and use the index from it`

osmo_jupyter/plot.py
jason-curtis: I'll do that

osmo_jupyter/plot.py
jason-curtis: it references the location of the annotation to the "canvas", not to the axis values. so 0.5 would be the middle no matter what the X values are.
I'm still fiddling with this a bit since the multiple-y-axes bug has resurfaced :( :(

osmo_jupyter/plot.py
jason-curtis: because otherwise it doesn't fix the bug.

setup.py
jason-curtis: yeah I'll send something out but the backwards incompatibilities turn out to be less of a big deal than they initially seemed. And this gets us a better jupyter interface (through real Widgets!)

osmo_jupyter/plot.py
jason-curtis: doing that!

osmo_jupyter/plot.py
jason-curtis: added a comment to that effect

osmo_jupyter/plot.py
jason-curtis: removed the requirement.

osmo_jupyter/plot.py
jason-curtis: I (mostly) figured it out. They key is how the `overlaying` argument works, and I think I mostly know what it means now.

PR #7, opened 2018-10-12
add ability to override kwargs that we are already using

jason-curtis:
For instance, this allows you to override `opacity=0.5`. Previously this blew up with an error about getting two values for the same key.

### DIFF
--- a/osmo_jupyter/plot.py
+++ b/osmo_jupyter/plot.py
@@ -60,7 +60,7 @@ def get_scatter(
     marker_overrides = marker_overrides or {}
     scatter_overrides = scatter_overrides or {}
 
-    return go.Scatter(
+    scatter_kwargs = dict(
         x=x_column_values,
         y=y_values,
         name=name,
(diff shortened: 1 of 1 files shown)


osmo_jupyter/plot.py
```diff
@@ -60,7 +60,7 @@ def get_scatter(
     marker_overrides = marker_overrides or {}
     scatter_overrides = scatter_overrides or {}
 
-    return go.Scatter(
+    scatter_kwargs = dict(
```
jaimemarijke, line 63: why did you use dict() instead of {}?

osmo_jupyter/plot.py
```diff
@@ -72,7 +72,10 @@ def get_scatter(
         },
         opacity=0.8,
         yaxis=f'y{y_axis_number}',
-        **scatter_overrides
+    )
+    scatter_kwargs.update(scatter_overrides)
```
jaimemarijke, line 76: can you use a splat to unpack `scatter_overrides` inside of the `scatter_kwargs` definition?

jaimemarijke (approved):

osmo_jupyter/plot.py
jason-curtis: simpler diff

osmo_jupyter/plot.py
jason-curtis: I tried but I still got the error. I was kinda surprised because I thought it had worked before...

PR #8, opened 2018-11-27
Switch to all-pytest test run, tool to re-add RGB columns to a DataFrame

jason-curtis:
asana ticket https://app.asana.com/0/819671808102776/858476395277064/f
I let myself scope-creep to the toxification to get this repo in line with our best practices.

### DIFF
--- a/osmo_jupyter/plot.py
+++ b/osmo_jupyter/plot.py
@@ -153,6 +153,26 @@ def _add_dicts(dicts):
     return output_dict
 
 
+def _get_colors_with_pattern(pattern):
+    ''' Insert 'r', 'g', and 'b' into a given pattern, replacing the "*" in the pattern
+
+    Args:
+        pattern: string with an asterisk ("*") in place of the color letter, eg. '*_mean'
+    Return:
+        list of strings, with each color inserted into the pattern
+
+    >>> _get_colors_with_pattern('*_mean')
+    ['r_mean', 'g_mean', 'b_mean']
+    '''
+    if '*' not in pattern:
+        raise ValueError(f'Pattern "{pattern}" does not contain an asterisk ("*")')
+
+    return [
+        pattern.replace('*', color)
+        for color in COLOR_CHANNELS
+    ]
+
+
 def get_rgb_columns(df, pattern, other_columns_to_include=None):
     ''' Get columns from the input DataFrame corresponding to colors, removing a shared prefix or suffix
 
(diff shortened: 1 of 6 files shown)


osmo_jupyter/plot.py
```diff
@@ -153,6 +153,26 @@ def _add_dicts(dicts):
     return output_dict
 
 
+def _get_colors_with_pattern(pattern):
```
jaimemarijke, line 156: nit: `get` seems like a slight misnomer here as it's really constructing a thing vs. getting a thing. 

What's your rationale for requiring the '*' vs just providing a postfix? e.g. _get_colors_with_pattern('_mean')? Are there valid cases where the color is in the middle of the string? If so, maybe update the example to use that.

osmo_jupyter/plot.py
```diff
@@ -168,19 +188,39 @@ def get_rgb_columns(df, pattern, other_columns_to_include=None):
...
     return df[all_desired_columns].rename(columns=rename_map)
 
 
+def add_rgb_columns(df_to_modify, rgb_df, pattern):
+    ''' Add rgb columns to a DataFrame (IN PLACE!) with a provided prefix or suffix pattern
+
+    eg.
+    >>> main_df = pd.DataFrame(..., columns=['Timestamp', ...])
+    >>> rgb_df = pd.DataFrame(..., columns=['r', 'g', 'b'])
+    >>> get_rgb_columns(main_df, rgb_df, '*_mean')
```
jaimemarijke:
```suggestion
    >>> add_rgb_columns(main_df, rgb_df, '*_mean')
```

osmo_jupyter/plot.py
```diff
@@ -168,19 +188,39 @@ def get_rgb_columns(df, pattern, other_columns_to_include=None):
...
     }
     return df[all_desired_columns].rename(columns=rename_map)
 
 
+def add_rgb_columns(df_to_modify, rgb_df, pattern):
+    ''' Add rgb columns to a DataFrame (IN PLACE!) with a provided prefix or suffix pattern
+
+    eg.
+    >>> main_df = pd.DataFrame(..., columns=['Timestamp', ...])
+    >>> rgb_df = pd.DataFrame(..., columns=['r', 'g', 'b'])
```
jaimemarijke: I don't get why this df is in this example

jaimemarijke: ohh. Is this the actual means? Maybe the name could be clearer then.

```
>>> rgb_means_df = pd.DataFrame(..., columns=['r', 'g', 'b'])
```

osmo_jupyter/plot.py
```diff
@@ -168,19 +188,39 @@ def get_rgb_columns(df, pattern, other_columns_to_include=None):
...
+    >>> get_rgb_columns(main_df, rgb_df, '*_mean')
+    >>> main_df
+    <DataFrame with columns ['Timestamp', ..., 'r_mean', 'g_mean', 'b_mean'] >
+
+    Args:
+        df_to_modify: pandas.DataFrame to have columns added
+        rgb_df: pandas.DataFrame with columns ['r', 'g', 'b'] to be
```
jaimemarijke: incomplete sentence. I think you mean

```
rgb_df: pandas.DataFrame with columns ['r', 'g', 'b'] to be added
```

osmo_jupyter/plot_test.py
```diff
@@ -0,0 +1,65 @@
...
+            'r': orig_df['r_abcd'],
+            'g': orig_df['g_abcd'],
+            'b': orig_df['b_abcd'],
+        })
+
+        pd.testing.assert_frame_equal(actual, expected)
+
+
+class TestAddRgbColumns:
+    rgb_df = pd.DataFrame({
```
jaimemarijke: nit: maybe change to `rgb_abcd_df`

jaimemarijke: also might be slightly clearer if you used a real term, like `*_mean` and `rgb_means_df`

jaimemarijke (approved):

osmo_jupyter/plot.py
jason-curtis: not really, it was mainly designed to support a prefix or a suffix, and the middle thing was just a side effect. we seem to have since landed on more suffix-y patterns since then. The `'*'` pattern a tad more user-friendly to me, and there's a slight concern for backwards compatibility too (with existing uses of `get_rgb_columns()`, so I'm leaning towards keeping this as-is.
if you still disagree I can change it.

osmo_jupyter/plot_test.py
jason-curtis: I changed it; didn't like the idea of implying a particular purpose of the `rgb_df` but I changed it to `rgb_df_to_add`

osmo_jupyter/plot.py
jason-curtis: done

jaimemarijke (approved):

osmo_jupyter/plot.py
jason-curtis: re: misnomer: huh, I often use `get_` as a prefix to verbify my functions. I do it to avoid `colors_with_pattern = colors_with_pattern()' aliasing. `construct` seems a little excessive, though. do you have a different suggestion?

PR #9, opened 2018-12-12
Add a `pick_file` function that prompts user for a filepath

jaimemarijke:
Code for [Separate tool to select ROIs](https://app.asana.com/0/819671808102776/901314508849883)

It turns out the only hard part of this ticket was how to select image to use for ROI selection. I thought maybe it would be best to just give technicians a file picker dialog, hence this function.

Here's how I'll demonstrate this in the jupyter snippet:
image

**Open questions**
* Should I require passing in the directory in `pick_file`?. The file picker works fine without a directory, but I just worry that could lead to accidentally selecting an image from the wrong folder.
* Should I pull together the three functions in the example snippet into a single public function for convenience? This function would probably live in `process_experiment` e.g:
```
def select_ROIs(raw_images_directory):
    jpeg_file_path_for_ROI_selection = pick_file(raw_images_directory)
    rgb_image_for_ROI_selection = raw.open.as_rgb(jpeg_file_path_for_ROI_selection)
    ROI_definitions = prompt_for_ROI_selection(rgb_image_for_ROI_selection)
    return ROI_definitions
```

**Testing**
Only manual testing in jupyter notebook

### DIFF
--- a/osmo_jupyter/__init__.py
+++ b/osmo_jupyter/__init__.py
@@ -1,3 +1,3 @@
 from .db_access import configure_database, load_calculation_details  # noqa: F401
 from .timezone import utc_series_to_local  # noqa: F401
-from .files import pick_file
+from .files import pick_file  # noqa: F401
--- a/osmo_jupyter/__init__.py
+++ b/osmo_jupyter/__init__.py
@@ -1,2 +1,3 @@
 from .db_access import configure_database, load_calculation_details  # noqa: F401
 from .timezone import utc_series_to_local  # noqa: F401
+from .files import pick_file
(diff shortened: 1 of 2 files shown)


jaimemarijke:
@jacob-r-g @valerie-osmobot feel free to chime in here :)

dubchuck (commented):
LGTM, but think that Technicians should own the approval.

jason-curtis (approved):
LGTM. As for a default directory, I think it makes sense to have that in the example code - probably pass in SYNC_DIRECTORY or whatever it is that has the standard value of `~/cosmobot-data-sets`.

PR #10, opened 2019-01-08
create get_node_temperature_data in osmo_jupyter

valerie-osmobot:
aka Valerie's first sorta real thing. also tests.

### DIFF
--- a/osmo_jupyter/db_access.py
+++ b/osmo_jupyter/db_access.py
@@ -162,14 +162,13 @@ def load_calculation_details(
     return calculation_details
 
 
-def get_node_temperature_data(start_time_local, end_time_local, node_ids, downsample_factor=1):
+def get_node_temperature_data(start_time_local, end_time_local, node_id, downsample_factor=1):
     ''' Load node temperature data only from the calculation_details table
 
     Args:
         start_time_local: string of ISO-formatted start datetime in local time, inclusive
         end_time_local: string of ISO-formatted end datetime in local time, inclusive
-        node_ids: iterable of node IDs to get data for
-            Default 20038 as it is the node set up in the BBB.
+        node_id: node ID to get data from
         downsample_factor: if this is a number, it will be used to select fewer rows.
             You should get *roughly* n / downsample_factor samples.
     Returns:
@@ -180,7 +179,7 @@ def get_node_temperature_data(start_time_local, end_time_local, node_ids, downsa
 
     raw_node_data = load_calculation_details(
         db_engine,
-        node_ids,
+        [node_id],
         start_time_local,
         end_time_local,
         downsample_factor
--- a/osmo_jupyter/db_access.py
+++ b/osmo_jupyter/db_access.py
@@ -160,3 +160,41 @@ def load_calculation_details(
     )
     connection.close()
     return calculation_details
+
+
+def get_node_temperature_data(start_time_local, end_time_local, node_ids, downsample_factor=1):
+    ''' Load node temperature data only from the calculation_details table
+
+    Args:
+        start_time_local: string of ISO-formatted start datetime in local time, inclusive
+        end_time_local: string of ISO-formatted end datetime in local time, inclusive
+        node_ids: iterable of node IDs to get data for
+            Default 20038 as it is the node set up in the BBB.
+        downsample_factor: if this is a number, it will be used to select fewer rows.
+            You should get *roughly* n / downsample_factor samples.
+    Returns:
+        a pandas.DataFrame of temperature data (°C) in local time from the node IDs provided.
+    '''
+
+    db_engine = configure_database()
+
+    raw_node_data = load_calculation_details(
+        db_engine,
+        node_ids,
+        start_time_local,
+        end_time_local,
+        downsample_factor
+    )
+
+    print(f'{len(raw_node_data)} rows retrieved.')
+
+    temperature_data = raw_node_data[raw_node_data['calculation_dimension'] == 'temperature']
+
+    temperature_data_only = pd.DataFrame({
+        'timestamp': timezone.utc_series_to_local(
+            temperature_data['create_date']
+        ),
+        'temperature': temperature_data['calculated_value'],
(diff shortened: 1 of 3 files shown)


osmo_jupyter/db_access_test.py
```diff
@@ -78,3 +82,48 @@ def test_to_aware_blows_up_if_timezone_provided():
...
+    mock_raw_node_data = pd.DataFrame({
+        'calculation_dimension': ['temperature', 'do'],
+        'calculated_value': [25.9, 7],
+        'create_date': ['01/01/2018 12:00', '01/01/2018 12:01'],
+        'other_things': ['ploop', 'more ploop'],
+    })
+
+    mock_load_calculation_details.return_value = mock_raw_node_data
+
+    mock_timezone_utc_series_to_local.side_effect = lambda x: pd.to_datetime(x)
```
jaimemarijke: It looks like this might just reproduce what's in the function you're mocking?
Instead, it would be better to return some kind of stub value, e.g.

`mock_timezone_utc_series_to_local.side_effect = lambda x: sentinel.local_datetime`
or
`mock_timezone_utc_series_to_local.side_effect = lambda x: x`
or
`mock_timezone_utc_series_to_local.side_effect = lambda x: f'{x}-converted-to-local'`

And then update line 119 accordingly.

osmo_jupyter/db_access_test.py
```diff
@@ -78,3 +82,48 @@ def test_to_aware_blows_up_if_timezone_provided():
...
+    return mocker.patch.object(timezone, 'utc_series_to_local')
+
+
+def test_get_node_temperature_data(
+    mock_configure_database,
+    mock_load_calculation_details,
+    mock_timezone_utc_series_to_local
+):
+    mock_raw_node_data = pd.DataFrame({
+        'calculation_dimension': ['temperature', 'do'],
```
jaimemarijke: might be good to test having more than one 'temperature' data point in here, so that the output series contains more than one value

jaimemarijke (commented):
A few comments. I'll let someone else give the approving review

osmo_jupyter/db_access.py
```diff
@@ -160,3 +160,41 @@ def load_calculation_details(
...
+
+
+def get_node_temperature_data(start_time_local, end_time_local, node_ids, downsample_factor=1):
+    ''' Load node temperature data only from the calculation_details table
+
+    Args:
+        start_time_local: string of ISO-formatted start datetime in local time, inclusive
+        end_time_local: string of ISO-formatted end datetime in local time, inclusive
+        node_ids: iterable of node IDs to get data for
+            Default 20038 as it is the node set up in the BBB.
```
jason-curtis: rly?
not seeing this implemented up in here

osmo_jupyter/db_access_test.py
```diff
@@ -78,3 +82,48 @@ def test_to_aware_blows_up_if_timezone_provided():
...
+
+
+@pytest.fixture
+def mock_configure_database(mocker):
+    mocker.patch.object(module, 'configure_database')
+
+
+@pytest.fixture
+def mock_load_calculation_details(mocker):
+    return mocker.patch.object(module, 'load_calculation_details')
```
jason-curtis, line 93: these separate mocks don't need to be in functions unless they're reused... maybe you should just do the mocking inside the test until someone else needs the same thing?

jason-curtis (approved):
👍 🚢 
only minor issues in the comments

osmo_jupyter/db_access.py
dubchuck: To add a little more, I believe that Jason is saying that the ```node_ids``` parameter should have a default arg similar to ```downsample=1```.

e.g. ```node_ids=[20038]```

dubchuck (commented):
Looking great, light review with one comment.

osmo_jupyter/db_access.py
jason-curtis: right. but then that's a magic number so maybe it should be a top-level constant
`BBB_TEMPERATURE_NODE_ID = 20038`
and then the default value `node_ids=[20038]`

thinking about it even a little bit more, does it really make sense for this function to take multiple node IDs? any information about what node it came from gets removed from the output, so the function would probably need some modification to be useful with multiple nodes. I think I'd vote for changing this to just take a single `node_id`.

osmo_jupyter/db_access_test.py
valerie-osmobot: yep, will do.

osmo_jupyter/db_access_test.py
valerie-osmobot: The function I'm mocking both converts a string to a datetime and changes its timezone. What does the first option in this comment do?

jason-curtis: it would just always return the same value. It would be more-straightforwardly implemented as `mock_timezone_utc_series_to_local.return_value = sentinel.local_datetime`

the mocked version of the function probably shouldn't just do the same thing as the original, otherwise why mock it? you can always just use the actual, original sub-function. The cost is that it adds more reasons that the test might fail, and thus you will have to change this test if the sub-function changes at all, but the benefit is that you're actually testing that the sub-function is being used correctly.

osmo_jupyter/db_access.py
valerie-osmobot: ooh, just say these. yeah, I decided it shouldn't default to 20038, changed that, and forgot to change documentation.

re: multiple nodes -- yeah, that makes sense.

osmo_jupyter/db_access_test.py
valerie-osmobot: okay, so per my discussion with Jason, I ended up not using the mock at all.

jason-curtis (approved):

osmo_jupyter/db_access_test.py
jaimemarijke: oops, I had her write it this way when I was thinking we'd write more than one test.

PR #11, opened 2019-01-22
Add functions to import spectrometer data and create intensity summary

jacob-r-g:
**Work**
- Created function to preprocess and import raw spectrometer txt file
- Created function to create intensity summary from imported spectrometer data 

**Testing**
- Added and ran new automated tests 

**ToDo**
- Manually test in a jupyter notebook


osmo_jupyter/mock_bad_spectrometer_data.txt
```diff
@@ -0,0 +1,3 @@
+2019-01-07 16:13:37.597000	1546906417597	-626	627	-163.15
```
jaimemarijke: We talked about this but forgot to do it: probably we should move these files into a new "text_fixtures" folder, e.g.:  `/osmo_jupyter/test_fixtures/mock_bad_spectrometer_data.txt`. (You'll need to update the test code that specifies where to find these files as well).

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
+import pandas as pd
+import io
+
+
+def _import_spectrometer_txt(spectrometer_data_file_path):
+    with open(spectrometer_data_file_path, 'r', newline=None) as infile:
```
jaimemarijke, line 11: Should add a comment above this line that explains why we had to do this: e.g.
`# The raw spectrometer file contains mixed "\r\n" and "\n" line endings. Open file with newline=None to ensure all line endings are "\n"`

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+        for line in infile:
+            if line.strip() == '>>>>>Begin Spectral Data<<<<<':
```
jaimemarijke: probably this string should be a named constant, e.g.

`SPECTRAL_DATA_IDENTIFIER = '>>>>>Begin Spectral Data<<<<<'`

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
+import pandas as pd
+import io
+
+
+def _import_spectrometer_txt(spectrometer_data_file_path):
+    with open(spectrometer_data_file_path, 'r', newline=None) as infile:
+        for line in infile:
```
jaimemarijke, line 13: this loop should also have a comment, e.g.
`# Read through file metadata until we find the marker that identifies the start of the tab-delimited data`

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+            if line.strip() == '>>>>>Begin Spectral Data<<<<<':
+                file_content_with_unix_endings = infile.read()
+                unformatted_spectrometer_data = pd.read_csv(
+                    io.StringIO(file_content_with_unix_endings),
```
jaimemarijke: could probably skip the temporary variable `file_content_with_unix_endings` and just have this line directly call `infile.read()`

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+                    sep='\t',
+                    lineterminator='\n',
+                )
+
+                return unformatted_spectrometer_data
+        raise ValueError('Spectrometer data not in expected format; check the file header')
+
+
+def _clean_up_spectrometer_data_header(unformatted_spectrometer_data):
+
```
jaimemarijke: extra newline?

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+        unformatted_spectrometer_data: a pandas df of original spectrometer data,
+            imported with _import_spectrometer_txt()
+
+    Return:
+        A pandas dataframe with a sensible header, and only one timestamp column
+    '''
+
+    timestamp_column, epoch_column = unformatted_spectrometer_data.columns[[0, 1]]
+    spectrometer_data_with_clean_header = unformatted_spectrometer_data.rename(
+        # timestamp column is in local time, epoch_time column is in UTC
```
jaimemarijke: nit: comments should start with capitals.
Also... it makes me slightly nervous to be using the local timestamp instead of the UTC one. I wonder if we should switch to using the UTC epoch time and convert to a timezone-aware datetime

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+        columns={
+            timestamp_column: 'timestamp',
+            epoch_column: 'epoch_time'
+        }
+    )
+    # Remove redundant timestamp column
```
jaimemarijke: nit: timestamp --> epoch_time

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+    spectrometer_data_with_clean_header.drop(['epoch_time'], axis=1, inplace=True)
+
+    return spectrometer_data_with_clean_header
+
+
+def _reformat_spectrometer_data(spectrometer_data_with_clean_header):
+    ''' Reformats a df of spectrometer data with many wavelength columns,to a more useful format with 2 columns
```
jaimemarijke: nit: "columns,to" --> "columns, to"

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+        spectrometer_data_file_path: file path to .txt file produce by oceanview software in the format
+            'Time Series (column data)'
+
+    Return:
+        A pandas dataframe, indexed by timestamp, with 2 columns: wavelength and intensity
+    '''
+    unformatted_spectrometer_data = _import_spectrometer_txt(spectrometer_data_file_path)
+    spectrometer_data_with_clean_header = _clean_up_spectrometer_data_header(unformatted_spectrometer_data)
+    reformatted_spectrometer_data = _reformat_spectrometer_data(spectrometer_data_with_clean_header)
```
jaimemarijke: nice, I like this sequential structure better - the named temporary variables help me understand the sequence of steps that are occurring here

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return reformatted_spectrometer_data
+
+
+def spectrometer_intensity_summary(reformatted_spectrometer_data, minimum_wavelength=340, maximum_wavelength=1035):
```
jaimemarijke: nit: the default kwargs here feel like "magic numbers" - as a reader I don't know why those particular values were chosen.
If they are arbitrary, I'd probably just remove them. Or if you want either bound to be optional, then set them to "None" and have the code interpret that as unbounded.
If they have some physical basis, e.g. they are the min and max values that it is possible to read on the spectrometer, you could save them as named constants, e.g.
```
MIN_SPECTROMETER_WAVELENGTH=340
MAX_SPECTROMETER_WAVELENGTH=1035
...
def spectrometer_intensity_summary(reformatted_spectrometer_data, minimum_wavelength=MIN_SPECTROMETER_WAVELENGTH, maximum_wavelength=MAX_SPECTROMETER_WAVELENGTH):
```

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Return:
+        A pandas df, indexed by timestamp, with 1 intensity column that summarizes (calculates mean) intensity
+            data over some spectrum
+    '''
+
+    relevant_wavelengths = reformatted_spectrometer_data['wavelength'].between(minimum_wavelength, maximum_wavelength)
+    filtered_spectrometer_data = reformatted_spectrometer_data[relevant_wavelengths]
+
+    # Question whether or not this should be mean. An integral may relate more closely to a useful unit
```
jaimemarijke: TODO: answer this question, or at least change the comment to describe the rationale for choosing "mean()" for now

osmo_jupyter/spectrometer_test.py
```diff
@@ -0,0 +1,192 @@
...
+        ],
+
+    }).set_index('timestamp')
+    pd.testing.assert_frame_equal(actual, expected_formatted_spectrometer_data)
+
+
+def test_import_and_format_spectrometer_data():
+    test_data_file_path = pkg_resources.resource_filename('osmo_jupyter', 'mock_spectrometer_data.txt')
+    actual = module.import_and_format_spectrometer_data(test_data_file_path)
+    expected_formatted_spectrometer_data = pd.DataFrame({
```
jaimemarijke: since the exact same DataFrame is used in three tests, it would probably be good to pull it out to a named module-level constant so that you don't have to define it multiple times, e.g. 

```
FORMATTED_SPECTROMETER_DF = pd.DataFrame({...})
...
assert actual == FORMATTED_SPECTROMETER_DF
```

osmo_jupyter/spectrometer_test.py
```diff
@@ -0,0 +1,192 @@
...
+            -546.47,
+            -163.15,
+            -79.00,
+            189.10,
+        ],
+    }).set_index('timestamp')
+
+    actual = module.spectrometer_intensity_summary(
+        sample_formatted_spectrometer_data,
+        minimum_wavelength=344,
```
jaimemarijke: depending on what you decide to do about the default kwargs in the actual code, I might recommend a "parameterized" test that exercises multiple code paths, e.g. including both max & min, including just one of max or min, including neither max nor min.

If you go that route, I can show you how to quickly update this test to do that.

jaimemarijke (commented):
A bunch of small comments. 
I'll someone else giving an Approving review since I paired with you on this

osmo_jupyter/mock_bad_spectrometer_data.txt
jason-curtis: +1

osmo_jupyter/mock_spectrometer_data.txt
```diff
@@ -0,0 +1,17 @@
...
+Electric dark correction enabled: true
+Nonlinearity correction enabled: false
+Boxcar width: 0
+XAxis mode: Wavelengths
+Number of Pixels in Spectrum: 3648
+>>>>>Begin Spectral Data<<<<<
+		344.05	1031.859	1032.175
+2019-01-07 16:13:37.597000	1546906417597	-626	627	-163.15
+2019-01-07 16:13:42.397000	1546906422397	-640.67	198.99	-79
+2019-01-07 16:13:47.496000	1546906427496	-546.47	-546.47	189.1
```
jason-curtis: as mentioned in person, plz confirm that this has mixed line endings even in git, since tools sometimes normalize them without telling you. maybe not a huge deal if they're not mixed, but it'd at least be good to know.

osmo_jupyter/spectrometer.py
jason-curtis: if you do this, it'd be nice to explicitly point out that it's redundant to the timestamp column
```suggestion
    # 'epoch_time' column is redundant to 'timestamp' column. Remove it
```

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+                    lineterminator='\n',
+                )
+
+                return unformatted_spectrometer_data
+        raise ValueError('Spectrometer data not in expected format; check the file header')
+
+
+def _clean_up_spectrometer_data_header(unformatted_spectrometer_data):
+
+    ''' Cleans up header of raw spectrometer data, and removes redundant timestamp column.
```
jason-curtis: nice one-liner :)

osmo_jupyter/spectrometer.py
jason-curtis: +1, I think mean is fine. I'd be tempted to upgrade the function to divide by the bin width to actually give the mean value per nanometer wavelength, but I think it's also fine to keep this as-is.

osmo_jupyter/spectrometer_test.py
```diff
@@ -0,0 +1,192 @@
+import pkg_resources
+from unittest.mock import sentinel
+
+import pandas as pd
+import pytest
+
+from . import spectrometer as module
+
+
+class TestImportSpectrometerData:
```
jason-curtis, line 48: nice, I love how simple and convincing these integration tests are 👍 👍

osmo_jupyter/spectrometer_test.py
```diff
@@ -0,0 +1,192 @@
...
+    expected_high_wavelength_intensity_summary = pd.DataFrame({
+        'timestamp': [
+            '2019-01-07 16:13:37.597000',
+            '2019-01-07 16:13:42.397000',
+            '2019-01-07 16:13:47.496000',
+        ],
+        'intensity': [
+            0.5,
+            -220.84,
+            -546.47,
```
jason-curtis: you could simplify this test and the fixture, and make the results more self-evidently correct, by using only 2 timestamps and using round numbers that are easier to check that they're obviously correct.

osmo_jupyter/spectrometer_test.py
jason-curtis: +1 to this, it looks like you're only actually testing `maximum_wavelength` here but I'm not 100% sure since I'm too lazy to check the math (see later comment re: using round numbers when reasonable).
also happy to help you parameterize this though you could also just search for examples where we use `pytest.mark.parametrize`(sic).

osmo_jupyter/mock_spectrometer_data.txt
jason-curtis: (you checked, they're still there, all good!)

jason-curtis (changes requested):
mostly looks great!
lots of small things, the big ones are to move the .txt fixtures to a `fixtures` or `test_fixtures` directory and to make sure you're testing all the kwargs in that last test case.

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
+import pandas as pd
+import io
+
+
+def _import_spectrometer_txt(spectrometer_data_file_path):
+    with open(spectrometer_data_file_path, 'r', newline=None) as infile:
+        for line in infile:
+            if line.strip() == '>>>>>Begin Spectral Data<<<<<':
+                file_content_with_unix_endings = infile.read()
```
dubchuck: I believe this variable name is a bit over-descriptive, ```file_content``` would probably suffice.

Also there probably is not a guarantee there would always be ```unix_endings```.

osmo_jupyter/spectrometer.py
dubchuck: +1

osmo_jupyter/spectrometer.py
```diff
@@ -0,0 +1,107 @@
...
+                return unformatted_spectrometer_data
+        raise ValueError('Spectrometer data not in expected format; check the file header')
+
+
+def _clean_up_spectrometer_data_header(unformatted_spectrometer_data):
+
+    ''' Cleans up header of raw spectrometer data, and removes redundant timestamp column.
+
+    Args:
+        unformatted_spectrometer_data: a pandas df of original spectrometer data,
```
dubchuck: ```_df``` suffix instead of ```_data```?

dubchuck (commented):
Mostly nits.

osmo_jupyter/mock_bad_spectrometer_data.txt
jacob-r-g: definitely makes sense! worried about the problem we ran into earlier where pytest looks for anything with the prefix 'test_' any other suggestions for names? maybe testing_fixtures?

osmo_jupyter/spectrometer.py
jacob-r-g: awesome. thanks!

osmo_jupyter/spectrometer.py
jacob-r-g: sounds good

osmo_jupyter/spectrometer.py
jacob-r-g: good call!

osmo_jupyter/spectrometer.py
jacob-r-g: make sense

osmo_jupyter/spectrometer.py
jacob-r-g: makes sense!

osmo_jupyter/spectrometer_test.py
jacob-r-g: nice! good call

osmo_jupyter/mock_bad_spectrometer_data.txt
jaimemarijke: I *think* that was because the file itself was prefixed with `test_` instead of `mock_`. I don't remember if pytest cares about folder names. I'd try the folder name `test_fixtures` and see if it works, and if not, just call it "fixtures"

osmo_jupyter/spectrometer_test.py
jaimemarijke: +1, I know it's annoying and you avoided changing the test file b/c of the line ending issue, but it's probably worth it to edit it to pick numbers that are easier to deal with

jacob-r-g: great point, and well taken. The rationale for not doing that is consistency. Nice to use the same test data for multiple tests where possible, which should be more evident now that this test function accepts the module level FORMATTED_SPECTROMETER_DF variable...

willing to change it if its too unclear though

osmo_jupyter/spectrometer.py
jacob-r-g: changed and added a comment for clarity on what's happening

osmo_jupyter/spectrometer.py
jacob-r-g: yes! great suggestion.

osmo_jupyter/spectrometer_test.py
jacob-r-g: have glanced through some examples used and still feel like i could use a bit of hand holding. would love to grab either one of you for a few minutes for tips on this

osmo_jupyter/spectrometer.py
jacob-r-g: nit: addressed. 

I hear ya on the local time. A redeeming factor for me is that the timestamp for the spectrometer data is set by the MSI which is set automatically... in other words, was just matching the protocol of our timestamps for our image data (image exif data set to local time by Pi) What do you think?

osmo_jupyter/spectrometer_test.py
```diff
@@ -0,0 +1,187 @@
...
+}).set_index('timestamp')
+
+
+class TestImportSpectrometerData:
+    def test_import_spectrometer_txt(self):
+        test_data_file_path = pkg_resources.resource_filename('osmo_jupyter',
+                                                              'test_fixtures/mock_spectrometer_data.txt'
+                                                              )
+        test_spectrometer_df = module._import_spectrometer_txt(test_data_file_path)
+        mock_spectrometer_df = pd.DataFrame({
```
jason-curtis: You're using "test" and "mock" here in non-standard ways which I found confusing.
More standard names for this would be `actual_spectrometer_df` and `expected_spectrometer_df` or even just `actual` and `expected`.

osmo_jupyter/spectrometer_test.py
jason-curtis: Makes sense. Ultimately that gets you something in making sure that all of the data sets are similarly realistic, but in the long term it usually ends up being worth it for test cases to be decoupled from one another and have their own tailored fixtures, since they can be much simpler -> easier to understand and maintain, and it makes it easier to test edge cases without updating a bunch of other tests. I think this is fine for now but I reserve the right to make other tests that don't use the same fixture in the future.

jason-curtis (approved):
just one super minor thing leftttt

osmo_jupyter/spectrometer_test.py
jacob-r-g: That makes sense. Thanks for explaining

dubchuck (approved):
Great work @jacob-r-g!  🚢 👍 ✈️ 🕺 💃

PR #13, opened 2019-03-05
Update documentation on pre-timestamp column

jaimemarijke:


osmo_jupyter/ysi.py
```diff
@@ -1,24 +1,39 @@
 import pandas as pd
 
 
-def _guard_index_is_datetime(ysi_data):
+def _guard_ysi_data_index_is_datetime(ysi_data):
     if not isinstance(ysi_data.index, pd.DatetimeIndex):
         raise ValueError(
-            '''
+            f'''
```
jason-curtis: unnecesary f-string?
```suggestion
            '''
```

jason-curtis (approved):

PR #14, opened 2019-03-07
Split up functions in plot.py into submodules in a plot folder

jaimemarijke:
**Ultimate Goal**
Add the "heatmapifier" function to the osmo-jupyter library.

Problem is that it would logically fit into the `plot.py` module, but that file is already really big. 

**Goal of this PR**
So in preparation, I first want to refactor that module into a `plot` subfolder and split up the functions into separate submodules. 

I preserved all of the original functions _except_ `subsample_for_plot` which I didn't find used anywhere within osmo-jupyter _or_ in any of our existing jupyter notebooks

**Testing**
* Moved existing automated tests into appropriate files in the `plot` subfolder
* Manually tested that I preserved the old import paths by leaving a `plot_test.py` file in the top-level and confirming that the old imports still worked (e.g. `from plot import get_scatter`)

**Remaining work**
* Land this first
* Spend 1-2 hours looking into whether there is a fast and cheap automatic documentation solution that would help technicians find where the various functions live, so that we maybe don't have to keep preserving old import paths
* Actually add the heatmapifier functions :P


osmo_jupyter/plot/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .layout import *  # noqa: F401, F403 unused wildcard imports
+from .scatter import *  # noqa: F401, F403 unused wildcard imports
+from .rgb_columns import *  # noqa: F401, F403 unused wildcard imports
+
+
+# Q: What code should live here?
+# A: Anything that is directly related to displaying a plot in a jupyter notebook
+
+# Q: Then why is the rgb_columns.py module here?
+# A: To preserve backwards compatibility, since it used to be in the ploy.py module
```
jason-curtis: lol

osmo_jupyter/plot/layout.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+AXIS_SIDES_BY_COLOR = {
+    'r': 'left',
+    'g': 'right',
+    'b': 'right',
+}
+
+
+def _axis_kwargs(axis_number, title, color=None, **y_axis_params):
```
jason-curtis: the wildcard import will include these in the __init__ and thus in notebooks.
Maybe this is OK but I think it would also be safe to do a pass for internal/external things and do explicit imports, just to make things a bit tighter.

osmo_jupyter/plot/rgb_columns.py
```diff
@@ -0,0 +1,69 @@
+from .constants import COLOR_CHANNELS
```
jason-curtis: I think it'd be reasonable to move this file out of this module. I doubt the breaking change will really bother anyone

jason-curtis (approved):
the one reservation I have here is around how tightly the layout and RGB scatter functions are logically coupled. Since they rely on a convention about which axes are which, it's kinda unfortunate that they ended up in separate files along with other stuff.

I don't think you really made this *worse* here, you just made it clear that my initial tightly-coupled architecture isn't super beautiful :stuck_out_tongue:

osmo_jupyter/plot/layout.py
jaimemarijke: I originally had it that way, but then I was worried that future people adding functions wouldn't think to make sure to import them here. 

But maybe that's actually better anyways, if we want scientists to switch to importing from the specific file and not the `plot` module? 

I don't have a strong opinion.

jason-curtis: +1. change it

jason-curtis (approved):

dubchuck (approved):

PR #15, opened 2019-03-08
Added heatmapify function + tests

jaimemarijke:
**Work**
* Copy over existing heatmap code from [Heatmap example](https://drive.google.com/drive/u/0/folders/1cZnlonCBPF1HdZw6myvkTZMma4eEdufI)
* Refactor a bit
* Write automated tests

**Discussion questions**
* I chose to leave out the code that slices an image by channel, and instead have the `heatmapify` function take just a 2D array. I could be convinced that the function should instead take `image` and `channel` params and do the slicing itself (that's how I originally wrote it). Opinions?
* I did a bit of refactoring on the existing code, but tried to limit it. Should I do more to make it cleaner / easier to read?

**Testing**
* Automated unit tests for `get_block_means_2d` and `cut_array2d`
* Automated smoke/integration test for `heatmapify`
* No automated tests for `display_heatmap`
* In a jupyter notebook, manually tested that heatmapify function works and is equivalent to old jupyter snippet


osmo_jupyter/plot/heatmap.py
```diff
@@ -0,0 +1,111 @@
...
+    # Divide an array into contiguous blocks of a given shape.
+    array_height, array_width = array.shape
+    block_height, block_width = block_shape
+
+    # Think, gridlines:
+    ycut = np.arange(0, array_height + 1, block_height)
+    xcut = np.arange(0, array_width + 1, block_width)
+
+    block_centers = [
+        ((ycut[i] + ycut[i+1] - 1) / 2, (xcut[j] + xcut[j+1] - 1) / 2)
```
jason-curtis: hmmmm, this would be kind of iffy in a code sample. any way you could make it better? 
maybe just more newlines?

---- musing below this point. feel free to ignore ----
I see now how `xcut` and `ycut` are actually not really like grid lines, since they are *on* the grid cells rather than next to them. if they were each increased by 0.5, then you wouldn't have to do this here. But then the actual indexing on line 35 would get mad since slice indices have to be integers. That thought cleared things up for me a bit but didn't make this line feel better... probably newlines are the main way to go.

osmo_jupyter/plot/heatmap_test.py
```diff
@@ -0,0 +1,165 @@
+import numpy as np
+import pytest
+
+from . import heatmap as module
+
+
+class TestCutArray2D:
+    def test_cut_odd_shape(self):
```
jason-curtis, line 10: sick test suite.
I like how you used integers to clarify your intent, but there's a weakness there too. Maybe you want one test case which makes sure that the order of elements isn't messed up within a block?

jason-curtis (changes requested):
test request is the only real "change requested". the other thing is minor
I did not read the code in too much detail - mostly assumed that things were not changed from the original implementation

osmo_jupyter/plot/heatmap.py
jaimemarijke: Bleh, yeah, I was really tempted to sink time into trying to make this clearer, but in the ~5 minutes I spent thinking of it I didn't come up with anything great. 

Added newlines - do they help at all?

jason-curtis (approved):

osmo_jupyter/plot/heatmap.py
jason-curtis: yeah it does help. I think it's good enough

osmo_jupyter/plot/heatmap.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Return:
+        (block_centers, blocks) tuple where
+            block_centers: a 1D array of 2-tuple coordinates for the centers of the sliced blocks
+            blocks: a 1D array of blocks of shape `block_shape`
+    '''
+    array_height, array_width = array.shape
+    block_height, block_width = block_shape
+
+    # Think, gridlines:
```
dubchuck, line 23: Nit: Could be a little bit more descriptive here.

osmo_jupyter/plot/heatmap.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
+import plotly.graph_objs as go
+import plotly.figure_factory as ff
+
+
+def cut_array2d(array, block_shape):
+    ''' Cuts up an array into blocks of shape `block_shape`. If `block_shape` doesn't factor perfectly into the
+        shape of the array, drops data from the right and bottom edges of the input `array`.
```
dubchuck, line 8: Not a strong opinion but it feels that if the data from the right and bottom edges needs to be dropped it should be done within separate method.

osmo_jupyter/plot/heatmap.py
```diff
@@ -0,0 +1,113 @@
...
+    # Construct a new 2D array that just contains the mean value of each block
+    block_means_2d = np.reshape(
+        block_means,
+        newshape=(blocks_height, blocks_width)
+    )
+
+    return block_means_2d
+
+
+def display_heatmap(array, display_decimals=2, title='Heatmap'):
```
dubchuck, line 73: Possible nit: ```array``` could be more descriptive?

osmo_jupyter/plot/heatmap.py
```diff
@@ -0,0 +1,113 @@
...
+        annotation_text=np.round(array, display_decimals),
+    )
+
+    heatmap.layout.yaxis.autorange = 'reversed'
+    heatmap.layout.title = title
+
+    return go.FigureWidget(heatmap)
+
+
+def heatmapify(array, block_shape=(50, 50), display_decimals=2, title='Heatmap'):
```
dubchuck: Same nit as above

dubchuck (approved):
A few questionable nits but LGTM

osmo_jupyter/plot/heatmap.py
```diff
@@ -42,17 +42,18 @@ def cut_array2d(array, block_shape):
...
 
 
-def get_block_means_2d(block_centers, blocks):
+def get_block_means_2d(block_centers, blocks, averaging_function=np.mean):
     ''' Given 1D arrays of block_centers and blocks, construct a 2D array of block means.
 
     Args:
         block_centers: a 1D array of 2-tuple coordinates for the centers of the sliced blocks
         blocks: a 1D array of 2D blocks
+        averaging_function: Optional (default=np.mean). Function to use to get the average of each block
```
jason-curtis: might want to mention that it will be called with a 2d array

jason-curtis (approved):

osmo_jupyter/plot/heatmap.py
jaimemarijke: I had that thought too, but couldn't come up with a better name. Maybe `array_to_heatmapify`?

PR #20, opened 2019-04-08
Librarify simulation code

jaimemarijke:
**Work**
Librarify functions for:
* Simulating the optical reading from a patch given DO and temperature
* Simulating the spatial ratiometric result from a sealed vs. unsealed patch
* Plotting surfaces of meshgrids of DO and temperature

**Testing**
* Wrote and ran automated tests
* Manually tested the plotting functions in a jupyter notebook

**TODO**
* Should I write a jupyter notebook example snippet?
* ~Update published docs~


osmo_jupyter/simulate/do_patch.py
```diff
@@ -0,0 +1,116 @@
...
+        TEMPERATURE_STANDARD_OPERATING_MAX,
+    )
+    fit_optical_reading_max = _estimate_optical_reading(
+        0,
+        TEMPERATURE_STANDARD_OPERATING_MIN,
+    )
+    fit_optical_reading_range = fit_optical_reading_max - fit_optical_reading_min
+
+    # TODO: fix magic number
+    do_partial_pressure = do_pct_sat * 1.6
```
jason-curtis: looks like `DO_MAX` isn't imported yet, but I believe this is what you'll want
```suggestion
    do_partial_pressure = do_pct_sat * DO_PARTIAL_PRESSURE_MMHG_AT_1ATM / DO_MAX
```

osmo_jupyter/simulate/do_patch.py
```diff
@@ -0,0 +1,116 @@
...
+        [
+            go.Scatter(
+                x=do_domain,
+                y=get_optical_reading_normalized(do_domain, temperature),
+                mode='lines',
+                line={
+                    'color': color_from_temperature(temperature),
+                    'width': 1,
+                },
+                name=f'T={temperature:d}',
```
jason-curtis: what does the `:d` do?

osmo_jupyter/simulate/do_patch_test.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    def test_normalized_reading_middlin_values_are_middlin(self, name, do, temperature):
+        assert 0 < module.get_optical_reading_normalized(do, temperature) < 1
+
+
+class TestGetRateConstant:
+    def test_get_rate_constant(self):
+        # Spot-check rate function with some actual numbers
+        actual = module._get_rate_constant(temperature_c=0, preexponential_factor=10, activation_energy=1)
+        assert actual == pytest.approx(9.995598)
```
jason-curtis: woah that's cool. didn't know `pytest.approx` was a thing. I might still reach for `np.testing.assert_almost_equal` first in the future though

osmo_jupyter/simulate/do_patch_test.py
```diff
@@ -0,0 +1,51 @@
...
+        'name, do_pct_sat, temperature_c, expected',
+        [
+            ('zero DO min temp', 0, TEMPERATURE_STANDARD_OPERATING_MIN, 1.727900),
+            ('zero DO max temp', 0, TEMPERATURE_STANDARD_OPERATING_MAX, 1.268441),
+            ('max DO min temp', 100, TEMPERATURE_STANDARD_OPERATING_MIN, 0.625286),
+            ('max DO max temp', 100, TEMPERATURE_STANDARD_OPERATING_MAX, 0.396912),
+        ]
+    )
+    def test_estimate_optical_reading(self, name, do_pct_sat, temperature_c, expected):
+        # Spot-check function with some actual numbers
```
jason-curtis:
```suggestion
        # Spot-check function with some actual numbers. This will of course change if we update the curve.
```

osmo_jupyter/simulate/do_ratiometric_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.fixture
+def mock_get_optical_reading_normalized(mocker):
+    return mocker.patch.object(module, 'get_optical_reading_normalized')
+
+
+class TestSimulateSpatialRatiometricReading:
+    def test_same_do_returns_1(self):
+        actual = module.simulate_spatial_ratiometric_reading(do=0, temperature=0, sealed_patch_do=0)
+        assert actual == 1
```
jason-curtis, line 15: ooh I like this. maybe we could use a few more that talk about what should happen in general if one DO is greater than the other?

osmo_jupyter/simulate/do_ratiometric_test.py
```diff
@@ -0,0 +1,54 @@
...
+            expected_unsealed_patch_call,
+            expected_sealed_patch_call
+        ])
+
+    def test_returns_ratio_of_unsealed_to_sealed(self, mock_get_optical_reading_normalized):
+        mock_optical_reading_for_unsealed_patch = 1
+        mock_optical_reading_for_sealed_patch = 2
+
+        # Set up the mock to return "1" the first time (when it is called for the *unsealed* patch)
+        # and "2" the second (when it is called for the *sealed* patch)
```
jason-curtis, line 50: yuck. probably fine though

osmo_jupyter/simulate/plot.py
```diff
@@ -0,0 +1,106 @@
...
+
+def plot_surface_of_do_and_temp(
+    fn_or_meshgrid_of_do_and_temp,
+    title,
+    z_axis_title,
+):
+    ''' Create a surface plot of a DO- and temperature-dependent function.
+
+    Args:
+        fn_or_meshgrid_of_do_and_temp: either a function which takes DO saturation and temperature as positional
```
jason-curtis: in response to your wise review comments, I think I did a better job of factoring this in a different notebook: basically I just made this function take a meshgrid, and added a second function `plot_surface_of_do_and_temp_fn()` which takes a function. I think that would help clear stuff up here too.

osmo_jupyter/simulate/plot.py
```diff
@@ -0,0 +1,106 @@
...
+)
+DO_DOMAIN = np.linspace(DO_MIN, DO_MAX, (DO_MAX - DO_MIN) + 1)
+
+# In this notebook, we'll use a standard meshgrid of temperature and DO defined based on the domains above.
+# Whenever you see a "meshgrid" passed around, it corresponds to these temperature and DO values.
+# Use "indexing='ij'" to orient the meshgrids correctly so that dimension 0 iterates over DO
+DO_MESHGRID, TEMPERATURE_MESHGRID = np.meshgrid(DO_DOMAIN, TEMPERATURE_DOMAIN, indexing='ij')
+
+
+def get_meshgrid_of_do_and_temp(fn_of_do_and_temp):
```
jason-curtis: we'll likely want this outside of plotting fairly often. I almost want a `.do_temp_meshgrid` module here, but don't feel that strongly about it.

osmo_jupyter/constants.py
```diff
@@ -1,3 +1,31 @@
+
+# Science
+IDEAL_GAS_CONSTANT_J_PER_MOL_K = 8.314  # J / (mol * K)
+IDEAL_GAS_CONSTANT_L_ATM_PER_MOL_K = 0.08205  # (L * atm) / (mol * K)
+
+DEGREES_CELSIUS_AT_ZERO_KELVIN = 273.15
+
+# Full range of possible DOs, in percent sat
```
jason-curtis:
```suggestion
# Full range of possible (in normal circumstances) DOs, in percent sat
```

jason-curtis (commented):
everything looks pretty good. TODOs look generally appropriate though you could cut scope on some of them if you wanted. 

There is a lot of code in here, and there are many slightly different ways to set up these interfaces (e.g. the choice to not include dark noise here could be problematic later). There's some risk that we'll regret some of these choices later, but then I remind myself that but this feels like as good start as any!

osmo_jupyter/simulate/do_ratiometric_test.py
jaimemarijke: can you think of any cases we could be sure of, independent of the implementation of get_optical_reading_normalized?

osmo_jupyter/simulate/plot.py
jaimemarijke: +1

osmo_jupyter/simulate/do_patch_test.py
jaimemarijke: oh right, I only think of `np.testing.assert_almost_equal` when I'm comparing arrays. I'll just switch it

osmo_jupyter/simulate/plot.py
jaimemarijke: ok

jason-curtis (approved):

PR #21, opened 2019-04-12
DRY up documentation about source data fit

jaimemarijke:
Jacob requested that the docstrings of various functions include details about the source data and fit.

What it looks like:
image

### DIFF
--- a/osmo_jupyter/simulate/do_patch.py
+++ b/osmo_jupyter/simulate/do_patch.py
@@ -7,8 +7,24 @@ from osmo_jupyter.plot.color_from_temperature import color_from_temperature
 from osmo_jupyter.simulate.do_and_temp_meshgrid import DO_DOMAIN, TEMPERATURE_DOMAIN
 
 
+FIT_SOURCE_DOCUMENTATION = '''
+---
+Details about the optical reading fit and source data:
+    Source data: Ocean Optics patch data.
+
+    Fit type: a version of the two-site model which uses arrhenius equations for temperature dependence.
+    (Note that there are a lot of ways this could be tweaked.)
+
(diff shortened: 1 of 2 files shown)


osmo_jupyter/simulate/do_patch.py
```diff
@@ -7,8 +7,24 @@
 from osmo_jupyter.simulate.do_and_temp_meshgrid import DO_DOMAIN, TEMPERATURE_DOMAIN
 
 
+FIT_SOURCE_DOCUMENTATION = '''
```
jason-curtis, line 13:
```suggestion
# Sharing docstring content - note that this has to be shared via mutating a function's `__doc__` attribute.
# This is because the mechanism that defines __doc__ in the first place doesn't support addition or f-strings.
# For more exciting detail, check out https://bugs.python.org/issue28739 .
FIT_SOURCE_DOCUMENTATION = '''
```

jason-curtis (approved):
Just one suggestion to reduce the level of surprise here. Otherwise, LGTM

osmo_jupyter/simulate/do_patch.py
```diff
@@ -7,8 +7,24 @@
...
+---
+Details about the optical reading fit and source data:
+    Source data: Ocean Optics patch data.
```
jacob-r-g: this is amazing. to be super stubborn and nit-picky…. would be nice if it mentioned something about tau. For example “Source data: Ocean Optics calibration data where fluorescence decay was measured. We expect fluorescent decay to be directly proportional to fluorescence intensity across DO and temperature values”

jaimemarijke: sure thing :)

PR #22, opened 2019-05-02
Add DO calibration tools to osmo-jupyter

jason-curtis:
[Asana ticket](https://app.asana.com/0/819671808102776/1115223552332519/f)

After much (probably an unneessary amount of ) heming and hawing, I decided not to put any of this stuff in process-experiment. Reasons were:

* the purpose of osmo-jupyter is to prevent copypasta in notebooks. most of this stuff is in line with that and will only be used from notebooks for the near and medium-term future
* process-experiment is more of a purpose-built machine for one job - processing experimental image data. When we want to make DO interpretation a part of that, we'll figure out the appropriate interfaces we want for that and then we can move some things over if we have to. I think it's also possible that we'll want to keep process_experiment as an image processing library and have DO processing in a different library altogether, since it's logically a different step in the data pipeline. Basically I want to ~punt~ delegate that organization to our smarter future selves.

If you'd like to see this stuff in action, make yourself a copy of the following notebook:
Calibration/Cosmobot/Calibration toolkit-Jason librarification scratch.ipynb
There's some pretty neat stuff that happens in there. It would be great if at least one of you could review the code that remains in that notebook in addition to this code. Happy to do an in-person review.

## TODO

* PR
* rebuild docs after PR approval
* promote the scratch notebook to be the "main" notebook in that folder after landing


osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+):
+    '''Formula predicting DO based on optical reading and temperature
+    This is the *reverse* of estimate_optical_reading_two_site_model_with_temperature.
+
+    Source:
+    McNeil equation 7 (p.142) with arrhenius temperature dependence of i0, k_sv1 and k_sv2
+    Note: Lehner and McNeil are predicting tau, we are predicting intensity (i)
+
+    Args:
+        do_and_temp: tuple of dissolved oxygen (% saturation) and temeprature (Deg C)
```
EvanSimpson: Too much copypasta here - this should describe `optical_reading_and_temp`.

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+    T is absolute temperature in kelvin
+    A is preexponential factor (a constant)
+    Ea is the activation energy
+    R is the universal gas constant
+    '''
+    ideal_gas_constant = IDEAL_GAS_CONSTANT_J_PER_MOL_K
+    temperature_kelvin = temperature_c + DEGREES_CELSIUS_AT_ZERO_KELVIN
+
+    # Kinda silly, but this scales the activation energy to be more friendly to the regression
+    activation_energy_scaling_factor = 10000
```
EvanSimpson, line 25: How was 10000 determined?

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+def estimate_optical_reading_two_site_model_with_temperature(
+        do_and_temp,
+        f,
+        A_i0,
+        E_i0,
+        A_k_sv1,
+        A_k_sv2,
+        E_k_sv
+):
+    ''' Two-site model fit including arrhenius equations for temperature dependence
```
EvanSimpson, line 40: What does two-site mean?

osmo_jupyter/simulate/do_patch.py
```diff
@@ -1,73 +1,20 @@
...
-    DEGREES_CELSIUS_AT_ZERO_KELVIN, IDEAL_GAS_CONSTANT_J_PER_MOL_K, DO_PARTIAL_PRESSURE_MMHG_AT_1ATM, DO_MAX
+    DO_PARTIAL_PRESSURE_MMHG_AT_1ATM, DO_MAX
 from osmo_jupyter.plot.color_from_temperature import color_from_temperature
 from osmo_jupyter.simulate.do_and_temp_meshgrid import DO_DOMAIN, TEMPERATURE_DOMAIN
 
 
-# Sharing docstring content - note that this has to be shared via mutating a function's `__doc__` attribute.
-# This is because the mechanism that defines __doc__ in the first place doesn't support addition or f-strings.
-# For more exciting detail, check out https://bugs.python.org/issue28739.
-FIT_SOURCE_DOCUMENTATION = '''
```
jason-curtis, line 13: I was a bit on the fence about pulling this documentation out, and could be convinced to reinstate something like it if necessary.

Rationale for removing it:

It's not as relevant since we're using our own data now, rather than OO data. Also since the fit equation is in another file, we would have been sharing these docs across 3 files and it already felt cumbersome at 2.

osmo_jupyter/simulate/do_patch.py
```diff
@@ -1,73 +1,20 @@
...
-
-    As part of this, we convert DO values to partial pressure to match the partial pressure unit used in the
-    legacy fit.
-
-    README for where this came from and with a lot of other useful info:
-    https://docs.google.com/document/d/1lBv0aumCDdCd0JDt9hPw6aIExyhNtGtN9RVu_5sQ5yA/edit
-'''
-
-
-def _get_rate_constant(temperature_c, preexponential_factor, activation_energy):
```
jason-curtis, line 30: moved all of the curve math to curve.py and replaced it with updated versions.

One weirdness: the fit that's in here is technically a quasiratiometric fit. I think it's probably close enough, but I could run a different fit on raw DO patch data if we think that'd be worth it.

setup.py
```diff
@@ -10,6 +10,7 @@
     url='https://www.github.com/osmosystems/python-jupyter-tools',
     packages=find_packages(),
     install_requires=[
+        'ipython',
```
jason-curtis, line 13: need this to directly call `display()`

jason-curtis (commented):
annotating

setup.py
EvanSimpson, line 13: I think you need to add sklearn to this as well. I had to install that to run the example notebook.

jason-curtis: That was a leftover import in the notebook. cleaning up now.

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+    R is the universal gas constant
+    '''
+    ideal_gas_constant = IDEAL_GAS_CONSTANT_J_PER_MOL_K
+    temperature_kelvin = temperature_c + DEGREES_CELSIUS_AT_ZERO_KELVIN
+
+    # Kinda silly, but this scales the activation energy to be more friendly to the regression
+    activation_energy_scaling_factor = 10000
+    exponent = -activation_energy * activation_energy_scaling_factor / (ideal_gas_constant * temperature_kelvin)
+
+    return preexponential_factor * np.power(np.e, exponent)
```
jaimemarijke: FYI I think `np.exp(exponent)` is the same as `np.power(np.e, exponent)`
At least, that's what I used in my math

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+        E_k_sv
+):
+    ''' Two-site model fit including arrhenius equations for temperature dependence
+    The form is derived from:
+    Lehner Eqn. 1.22 with Arrhenius temperature dependence of i0, k_sv1 and k_sv2
+    as described by McNeil p.143 (though they do a more roundabout thing with β, it's equivalent)
+    Note: Lehner and McNeil are predicting tau, we are predicting intensity (i)
+
+    Args:
+        do_and_temp: tuple of dissolved oxygen (% saturation) and temeprature (Deg C)
```
jaimemarijke:
```suggestion
        do_and_temp: tuple of dissolved oxygen (% saturation) and temperature (Deg C)
```

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+        f: fraction of fluorophores in site 1 vs. site 2
+            should always be less than 1.
```
jaimemarijke:
```suggestion
            should always be in the range [0, 1]
```

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+        A_i0: Arrhenius preexponential parameter for unquenched fluorescence
```
jaimemarijke: nit: elsewhere you called this "preexponential factor" not "preexponential parameter". would be nice to be consistent (same for descriptions of A_k_sv1 and A_k_sv2)

jaimemarijke: and same for the docstrings of `estimate_do_two_site_model_with_temperature`

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+        E_i0: Arrhenius activation energy (divided by 10000 for fit friendliness) for unquenched fluorescence
+        A_k_sv1: Arrhenius preexponential parameter for stern-volmer fluorescence quenching model affected by oxygen
+            in site 1
+        A_k_sv2: Arrhenius preexponential parameter for stern-volmer fluorescence quenching model affected by oxygen
+            in site 2
+        E_k_sv: Arrhenius activation energy (divided by 10000 for fit friendliness) stern-volmer fluorescence
+            quenching model
+
+    Returns:
+        Estimate of dissolved oxygen
```
jaimemarijke: what units? % sat?

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+    optical_reading, temperature = optical_reading_and_temp
+    i = optical_reading
+
+    i0 = _get_arrhenius_rate(temperature, A_i0, E_i0)
+    k_sv1 = _get_arrhenius_rate(temperature, A_k_sv1, E_k_sv)
+    k_sv2 = _get_arrhenius_rate(temperature, A_k_sv2, E_k_sv)
+
+    sqrt_term = (
+        (i0 / i) ** 2 * (
+            -2 * f * k_sv1 ** 2
```
jaimemarijke: I think this would be slightly easier to read if you put parenthesis around all the exponents, e.g.
`-2 * f * (k_sv1 ** 2)`

I know that the computer knows to group them that way, but it's hard for my brain to do it.

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+        estimate_do_fn: function of ((optical reading, temperature), *fit params) which returns an estimate of DO % sat
+        initial_fit_params: fit parameters to seed the curve fit. Pass None to skip initialization
+
+    Returns:
+        optimized fit parameters
+
+    Raises:
+        Various errors from scipy.optimize.curve_fit if a fit cannot be found.
+    '''
+    fit_params, pcov_dO = curve_fit(
```
jaimemarijke: nit: `pcov_dO` could maybe have a better name? no idea what it is (well, other than I happen to know the return signature of `curve_fit`). unless `pcov` is a well known term in curve-fitting. If it's referencing DO, then it should at least be `_DO` instead of `_dO`

jaimemarijke: Though since it looks like it's unused, I suppose it could also just be `_`

osmo_jupyter/calibration/do/curve.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def get_optimal_DO_fit_params(
+        training_data,
+        estimate_do_fn=estimate_do_two_site_model_with_temperature,
+        initial_fit_params=WORKING_FIT_PARAMS
+):
+    ''' Optimize fit parameters for a DO fit
+    Args:
+        training_data: DataFrame of observations with 'SR reading' and 'Temperature (C)' columns
```
jaimemarijke:
```suggestion
        training_data: DataFrame of observations with 'SR reading', 'Temperature (C)', and 'DO (% sat)' columns
```

osmo_jupyter/calibration/do/prep.py
```diff
@@ -0,0 +1,71 @@
...
+    # Pull out some choice MSORMs
+    r_msorms = all_camera_data.set_index(['timestamp', 'ROI']).unstack()['r_msorm'].reset_index()
+
+    # Join with YSI data and prep for calibration
+    desired_ysi_columns = [
+        'Temperature (C)',
+        'Dissolved Oxygen (%)'
+    ]
+    msorms_and_ysi_data = join_nearest_ysi_data(r_msorms, ysi_data[desired_ysi_columns])
+    msorms_and_ysi_data.head()
```
jaimemarijke: vestigial? I assume this won't display in jupyter since it's called in the middle

osmo_jupyter/calibration/do/prep.py
```diff
@@ -0,0 +1,71 @@
+import pandas as pd
+
+from osmo_jupyter.ysi import join_nearest_ysi_data
+
+
+def prep_calibration_data(
+    ysi_data_filepath,
+    camera_data_filepath,
```
jaimemarijke: nit: here and elsewhere I think `image` is a slightly more appropriate term than `camera`

osmo_jupyter/calibration/do/prep_test.py
```diff
@@ -0,0 +1,115 @@
...
+            'irrelevant': 'other stuff'
+        }
+    ]).to_csv(camera_data_filepath)
+    return camera_data_filepath
+
+
+class TestPrepCalibrationData:
+
+    def test_rounds_ysi_temperature_data(self, ysi_data_filepath, camera_data_filepath):
+        # Note: this test requires the fixture to have non-integer temperatures in the YSI data
```
jaimemarijke: I'm not sure I follow why - the test will still pass if the YSI data starts with integers right? It's just that then it's not testing anything?

osmo_jupyter/calibration/do/curve.py
jason-curtis: added a section here: https://docs.google.com/document/d/1VDIn7b9xTZeuvdWRiF6P1HvRzde16HFVGKPHJrQfQGE/edit#heading=h.46ihesi06pxr

osmo_jupyter/calibration/do/curve.py
jason-curtis: improved the comment here - take a look

osmo_jupyter/calibration/do/prep_test.py
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    'timestamp': pd.Timestamp('2019-01-01 01:02:11'),
+                    'Temperature (C)': 35.0,
+                    'DO (% sat)': 1.0,
+                    'DO patch reading': 0.6,
+                    'Reference patch reading': 0.09,
+                    'SR reading': 6.666666666666667
+                }
+            ],
+            columns=expected_columns
```
jaimemarijke, line 112: I think you may have explained this before - why do you have to be explicit about the column names here? Won't they get picked up from the dictionary keys?

setup.py
jason-curtis: shouldn't be necessary anymore.

osmo_jupyter/calibration/do/prep_test.py
jaimemarijke: Also, should this `expected_calibration_data` be explicit about the df index?

osmo_jupyter/calibration/do/curve_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TestGetArrheniusRate:
+    def test_get_arrhenius_rate(self):
+        # Spot-check rate function with some actual numbers
+        actual = module._get_arrhenius_rate(
+            temperature_c=0,
+            preexponential_factor=10,
+            activation_energy=1e-4
+        )
+        np.testing.assert_almost_equal(actual, 9.9956, decimal=3)
```
jaimemarijke, line 52: dangit I just realized I forgot that this existed again and used pytest.approx in my own code.

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Returns:
+        Coefficient of determination - a measure of how much of the variation in actual data
+        is accounted for in the model
+    '''
+
+    sum_of_squares_of_residuals = ((actual - predicted) ** 2).sum()
+    total_sum_of_squares = ((actual - np.mean(actual)) ** 2).sum()
+
+    return 1 - (sum_of_squares_of_residuals / total_sum_of_squares)
```
jaimemarijke, line 30: I this just a thing and this is how it's defined?

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+        predicted: Y data points provided by a model
+        actual: Y data points actually observed
+
+    Returns:
+        Coefficient of determination - a measure of how much of the variation in actual data
+        is accounted for in the model
+    '''
+
+    sum_of_squares_of_residuals = ((actual - predicted) ** 2).sum()
+    total_sum_of_squares = ((actual - np.mean(actual)) ** 2).sum()
```
jaimemarijke, line 28: I don't think I understand why this is `total_sum_of_squares`.

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+    Params:
+        actual: actual values
+        predicted: values predicted by a model, corresponding to actual values
+        p: number of explanatory variables in the model
+    Returns:
+        adjusted r-squared value
+    '''
+    r_squared = _get_r_squared(actual, predicted)
+    n = len(actual)
+    return 1 - (1 - r_squared) * (n - 1) / (n - p - 1)
```
jaimemarijke, line 52: is the `(1 - r_squared)` term definitely different from the `1 - ..` term in `_get_r_squared`?

jaimemarijke: (I confirm that this equation looks exactly like the one you referenced from wikipedia, as long as the r_squared terms are the same)

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+            },
+            yaxis={
+                'title': 'DO error (predicted - actual)<br>(% saturation)',
+            },
+            hovermode='closest',
+        )
+    )
+
+
+def plot_calibration(
```
jaimemarijke, line 96: nit: this function is prefixed with `plot_` but the one above is postfixed with `_plot` -

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+    optical_reading = sensor_data['SR reading']
+    DO = sensor_data['DO (% sat)']
+    temperature = sensor_data['Temperature (C)']
+
+    predicted_do = estimate_do_fn(
+        (optical_reading, temperature),
+        *fit_params
+    )
+
+    p = len(fit_params)
```
jaimemarijke, line 126: nit: could just one-line this as a kwarg in 121

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+        p: number of explanatory variables in the model
+    Returns:
+        adjusted r-squared value
+    '''
+    r_squared = _get_r_squared(actual, predicted)
+    n = len(actual)
+    return 1 - (1 - r_squared) * (n - 1) / (n - p - 1)
+
+
+def _get_fit_parameters(fn):
```
jaimemarijke:
```suggestion
def _get_fit_parameter_names(fn):
```

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+        title: Title of fit, to use in plots
+        estimate_do_fn: Function of ((optical reading, temperature), *fit_params) which provides a DO estimate in % sat
+        estimate_optical_reading_fn:
+            Function of ((DO, temperature), *fit_params) which provides an optical reading estimate
+
+    Returns:
+        None. Produces plots and printed data frame
+    '''
+    optical_reading = sensor_data['SR reading']
+    DO = sensor_data['DO (% sat)']
```
jaimemarijke: nit: maybe call this `measured_do`?

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+    do_error = predicted_do - DO
+    max_do_error_index = do_error.abs().idxmax()
+    max_do_error = do_error[max_do_error_index]
+    worst_error_temperature = temperature[max_do_error_index]
+    worst_error_correct_do = DO[max_do_error_index]
+    worst_error_SR = optical_reading[max_do_error_index]
+
+    temperatures_text = [f'T={T}' for T in temperature]
+    print(dedent(f'''
+    Fit params: {{ {fit_params_formatted} }}
```
jaimemarijke, line 147: what do the `{{ .. }}` do here?

jaimemarijke (commented):
Partial review -didn't quite finish before mtg

osmo_jupyter/simulate/do_patch.py
jaimemarijke: I'd say just check with Jacob - he was the one who requested this doc in the first place

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+            text=temperatures_text,
+        ),
+    ]
+
+    error_lines = [
+        go.Scatter(
+            x=[point_do, point_predicted_do],
+            y=[point_optical_reading, point_optical_reading],
+            line=dict(
+                color='red',  # color_from_temperature(point_temperature),
```
jaimemarijke: vestigial?

osmo_jupyter/plot/do_calibration.py
```diff
@@ -0,0 +1,230 @@
...
+from typing import List, Callable
+
+from IPython.core.display import display
+import numpy as np
+import pandas as pd
+from plotly import graph_objs as go
+
+from osmo_jupyter.constants import TEMPERATURE_STANDARD_OPERATING_MIN, TEMPERATURE_STANDARD_OPERATING_MAX
+from osmo_jupyter.plot.color_from_temperature import color_from_temperature
+from osmo_jupyter.simulate.do_and_temp_meshgrid import DO_DOMAIN
```
jaimemarijke, line 12: I wonder if this should live here and be imported in simulate?

osmo_jupyter/simulate/do_patch_test.py
```diff
@@ -29,21 +29,14 @@ def test_normalized_reading_middlin_values_are_middlin(self, name, do, temperatu
...
         'name, do_pct_sat, temperature_c, expected',
         [
-            ('zero DO min temp', 0, TEMPERATURE_STANDARD_OPERATING_MIN, 1.7279),
-            ('zero DO max temp', 0, TEMPERATURE_STANDARD_OPERATING_MAX, 1.2684),
-            ('max DO min temp', 100, TEMPERATURE_STANDARD_OPERATING_MIN, 0.6253),
-            ('max DO max temp', 100, TEMPERATURE_STANDARD_OPERATING_MAX, 0.3969),
+            ('zero DO min temp', 0, TEMPERATURE_STANDARD_OPERATING_MIN, 4.4012),
+            ('zero DO max temp', 0, TEMPERATURE_STANDARD_OPERATING_MAX, 3.4646),
+            ('max DO min temp', 100, TEMPERATURE_STANDARD_OPERATING_MIN, 1.2577),
+            ('max DO max temp', 100, TEMPERATURE_STANDARD_OPERATING_MAX, 0.9995),
```
jaimemarijke, line 39: huh, did these numbers really change by that much between fits of OO data and our data?

jaimemarijke (approved):
Overall looks good!

osmo_jupyter/simulate/do_patch.py
jason-curtis: @jacob-r-g wanna chime in?

osmo_jupyter/calibration/do/curve.py
jason-curtis: thanks, that version feels more l33t and I will use it

osmo_jupyter/calibration/do/curve.py
jason-curtis: yes! and if you can figure out a way to remove like 2 more characters from these description I can remove all those darned line breaks too

osmo_jupyter/calibration/do/curve.py
jason-curtis: it's not really opinionated... depends on the parameters. But I will inject opinionation.

osmo_jupyter/calibration/do/curve.py
jason-curtis: oh yeah, those were more readable before I added the spaces to appease the linter! adding parens.

osmo_jupyter/calibration/do/curve.py
jason-curtis: yeah changing to _. I just used the name from curve_fit and didn't underscore it out like I should have when librarifying. thanks for pointing it out

osmo_jupyter/calibration/do/prep.py
jason-curtis: yep

osmo_jupyter/calibration/do/prep.py
jason-curtis: changing it

osmo_jupyter/calibration/do/prep_test.py
jason-curtis: right.

jason-curtis: updating to `# Note: test assumes the YSI data fixture has non-integer temperatures. Otherwise the test does nothing.`

osmo_jupyter/calibration/do/prep_test.py
jason-curtis: I had to specify the columns to get the ordering right :/
The index is just default numbering.

osmo_jupyter/calibration/do/curve_test.py
jason-curtis: pytest.approx is good too

osmo_jupyter/plot/do_calibration.py
jason-curtis: not sure I know what you mean, but surprisingly there doesn't seem to be a handy function for this in numpy... I had pulled in scikit-learn just for this but it seemed like overkill for a 3-line function.

osmo_jupyter/plot/do_calibration.py
jason-curtis: it's a [thing](https://en.wikipedia.org/wiki/Total_sum_of_squares). I agree that it's kind of a bad name. Having a hard time coming up with a better one that's not stupidly long though. `denominator`?

osmo_jupyter/plot/do_calibration.py
jason-curtis: yeah it's the same.

osmo_jupyter/plot/do_calibration.py
jason-curtis: and if you're asking if this is a thing, it's a very common statistic used for measuring fit quality.

osmo_jupyter/plot/do_calibration.py
jason-curtis: one of them is a noun (returns a plot), the other is a verb (displays a plot).

osmo_jupyter/plot/do_calibration.py
jason-curtis: :+1:

jason-curtis: oh actually I use it again on line 146

osmo_jupyter/plot/do_calibration.py
jason-curtis: :+1:

osmo_jupyter/plot/do_calibration.py
jason-curtis: in f-string land, those are escaped curlies
the result is something like `Fit params: { 'x': 12, 'y': 234.66 }
I have to do the formatting myself instead of just dropping the dict in there because I want to control the number formatting

osmo_jupyter/plot/do_calibration.py
jason-curtis: yep

osmo_jupyter/plot/do_calibration.py
jason-curtis: yeahhh I think it fits equally badly in both places. what's your case for moving it here?

osmo_jupyter/simulate/do_patch_test.py
jason-curtis: yeahhhh as noted in the description, it's probably largely because this is a spatial ratiometric curve and the old one was measuring tau. I could fit the new curve to a raw patch curve but I think once it's normalized it won't matter much.

osmo_jupyter/calibration/do/curve.py
jason-curtis: removed some other words, hopefully not too many.

EvanSimpson (approved):

PR #26, opened 2019-06-07
Add uniform dataframe sampler

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1125039984561184)

Adds a helper function to sample a given DataFrame uniformly across n-columns. Useful for balancing an ML dataset in e.g. DO and temperature values, with minimal lost data points.

### DIFF
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -23,13 +23,16 @@ def uniform(dataframe, columns_and_bin_counts):
     )
 
     # Combine all bins for each row into one series of n-dimensional bins
-    bins = binned.apply(tuple, 1)
+    combined_bins = binned.apply(tuple, axis=1)
 
     # Take samples from each bin
-    samples_per_bin = bins.value_counts().min()
-    samples_index = bins.groupby(bins).apply(
+    samples_per_bin = combined_bins.value_counts().min()
+    samples_index = combined_bins.groupby(
+        combined_bins,
+        group_keys=False
+    ).apply(
         lambda bin_group:
             bin_group.sample(samples_per_bin)
-    ).index.levels[1].values
+    ).index
 
     return dataframe.loc[samples_index]
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -1,37 +1,35 @@
-import numpy as np
 import pandas as pd
 
 
-def uniform(dataframe, columns, bin_counts):
+def uniform(dataframe, columns_and_bin_counts):
     '''
     Uniformly sample a DataFrame across n-dimensions.
 
     Args:
         dataframe: A pandas DataFrame
-        columns: A list of columns names to sample uniformly
-        bin_counts: A list of the number of bins over which to distribute each column
-        Number of bins are assigned positionally to their respective column
+        columns_and_bin_counts: A dictionary of columns names and the number of bins
+            over which to distribute each column
 
     Returns:
         A pandas DataFrame with a uniform distribution across the specified columns
     '''
 
-    # Create n-dimensional histogram of requested columns
-    histogram = pd.DataFrame(
-        [
-            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
-            for i in range(0, len(columns))
-        ],
-    ).transpose()
-    histogram.index = dataframe.index.copy()
-    # Combine all bins into one n-dimensional bin
-    histogram['bin'] = pd.Categorical(histogram.apply(tuple, 1))
+    # Create bins for each requested columns
+    binned = pd.DataFrame(
+        {
+            column: pd.cut(dataframe[column], bin_count)
+            for column, bin_count in columns_and_bin_counts.items()
+        },
+    )
 
-    max_samples = histogram['bin'].value_counts().min()
-    samples = [
-        histogram[histogram['bin'].astype(object) == bin].sample(max_samples).index.values
-        for bin in histogram['bin'].unique()
-    ]
+    # Combine all bins for each row into one series of n-dimensional bins
+    bins = binned.apply(tuple, 1)
 
-    index = np.array(samples).flatten()
-    return dataframe.loc[index]
+    # Take samples from each bin
+    samples_per_bin = bins.value_counts().min()
+    samples_index = bins.groupby(bins).apply(
+        lambda bin_group:
+            bin_group.sample(samples_per_bin)
+    ).index.levels[1].values
+
+    return dataframe.loc[samples_index]
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -1,0 +1,37 @@
+import numpy as np
+import pandas as pd
+
+
+def uniform(dataframe, columns, bin_counts):
+    '''
+    Uniformly sample a DataFrame across n-dimensions.
+
+    Args:
+        dataframe: A pandas DataFrame
+        columns: A list of columns names to sample uniformly
+        bin_counts: A list of the number of bins over which to distribute each column
+        Number of bins are assigned positionally to their respective column
+
+    Returns:
+        A pandas DataFrame with a uniform distribution across the specified columns
+    '''
+
+    # Create n-dimensional histogram of requested columns
+    histogram = pd.DataFrame(
+        [
+            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
+            for i in range(0, len(columns))
+        ],
+    ).transpose()
+    histogram.index = dataframe.index.copy()
+    # Combine all bins into one n-dimensional bin
+    histogram['bin'] = pd.Categorical(histogram.apply(tuple, 1))
+
+    max_samples = histogram['bin'].value_counts().min()
+    samples = [
+        histogram[histogram['bin'].astype(object) == bin].sample(max_samples).index.values
+        for bin in histogram['bin'].unique()
+    ]
+
+    index = np.array(samples).flatten()
+    return dataframe.loc[index]
--- a/osmo_jupyter/sample_test.py
+++ b/osmo_jupyter/sample_test.py
@@ -28,8 +28,10 @@ class TestSampleUniform(object):
             },
         )
 
-        expected_distribution = np.full(num_bins, 1 / num_bins)
+        # Use normalized value_counts to get the distribution of values in column 'one'
         actual_distribution = output_dataframe['one'].value_counts(normalize=True).values
+        # Create an array of expected value count distribution in column 'one'
+        expected_distribution = np.full(num_bins, 1 / num_bins)
 
         np.testing.assert_almost_equal(actual_distribution, expected_distribution)
 
@@ -43,10 +45,12 @@ class TestSampleUniform(object):
             },
         )
 
+        # Use normalized value_counts to get the distribution of values in column 'one'
+        actual_distribution = output_dataframe['one'].value_counts(normalize=True).values
+        # Expect the number of bins to saturate to the number of unique values instead of
+        # creating a large number of empty bins between max and min values in the column
         unique_values_count = len(self.small_test_dataframe['one'].unique())
         expected_distribution = np.full(unique_values_count, 1 / unique_values_count)
-
-        actual_distribution = output_dataframe['one'].value_counts(normalize=True).values
 
         np.testing.assert_almost_equal(actual_distribution, expected_distribution)
 
@@ -55,6 +59,7 @@ class TestSampleUniform(object):
         num_columns = 10
         columns = [f'column_{n}' for n in range(0, num_columns)]
         columns_to_balance = columns[0:4]
+
         input_dataframe = pd.DataFrame(
             np.random.rand(10000, num_columns),
             columns=columns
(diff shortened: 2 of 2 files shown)


osmo_jupyter/sample_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TestSampleUniform(object):
+    small_test_dataframe = pd.DataFrame(
+        [
+            (1, 1), (1, 2), (1, 3), (1, 4),
+            (2, 1), (2, 2), (2, 3), (2, 4),
+            (3, 1), (3, 2), (3, 3), (3, 4),
+            (4, 1), (4, 2), (4, 3), (4, 4),
+
+            (1, 1), (1, 1), (1, 2), (1, 2),
```
jason-curtis, line 16: this break implies something interesting is going on, but I would have grokked it slightly more quickly with a comment

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+    '''
+
+    # Create n-dimensional histogram of requested columns
+    histogram = pd.DataFrame(
+        [
+            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
+            for i in range(0, len(columns))
+        ],
+    ).transpose()
+    histogram.index = dataframe.index.copy()
```
jason-curtis: I don't think you should have to set the index yourself - it should just carry through.

Alternatively, you could tighten the whole thing up a bit by taking `columns_and_bin_counts` as one dict - then you could iterate over them together and eliminate a class of edge cases here. I'm a fan of keeping related data in a data structure that relates them rather than giving them the opportunity to not line up. If you did that and then instantiated the DataFrame using a dict of `{'column_name': pd.cut(...)}` then you wouldn't have to transpose either because I think pandas assumes the right thing in that case.

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+    # Combine all bins into one n-dimensional bin
+    histogram['bin'] = pd.Categorical(histogram.apply(tuple, 1))
+
+    max_samples = histogram['bin'].value_counts().min()
+    samples = [
+        histogram[histogram['bin'].astype(object) == bin].sample(max_samples).index.values
```
jason-curtis: why is this `.astype(object)` cast necessary?

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+    # Create n-dimensional histogram of requested columns
+    histogram = pd.DataFrame(
+        [
+            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
+            for i in range(0, len(columns))
+        ],
+    ).transpose()
+    histogram.index = dataframe.index.copy()
+    # Combine all bins into one n-dimensional bin
+    histogram['bin'] = pd.Categorical(histogram.apply(tuple, 1))
```
jason-curtis: There are some details I'm finding hard to follow here: Why do you need the additional `1` here? Why do you need to use `pd.Categorical` instead of just a `Series`?
Also, is there a need to attach this to the `histogram` at all? It might be easier to follow if this were just `combined_bin = pd.Categorical(...)` rather than `histogram['bin']`.

osmo_jupyter/sample_test.py
```diff
@@ -0,0 +1,71 @@
...
+        ],
+        columns=['one', 'two']
+    )
+
+    def test_uniform_in_one_dimension(self):
```
jason-curtis: I'm finding it slightly hard to grok what you're looking for in each test case as quickly as I expect to. Couple of thoughts on what might help:
1. In the title, have a quick explanation of not just the input data, but what you're looking for in each test case. Something like `test_sampling_on_one_dimension__flattens_desired_distribution`

2. separate out the `expected` value from the actual value on a different line - this gives the reader just a teense less cognitive overload when trying to understand the final `assert`. and when you're doing anything remotely clever in your assert (like you kind of are here) strongly consider having a comment about what it is you are asserting, esp. if you haven't found a way to express it in your test case title.

osmo_jupyter/sample_test.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+import pandas as pd
+
+import osmo_jupyter.sample as module
+
+
+class TestSampleUniform(object):
```
jason-curtis, line 7: you could add at least one test case that makes sure that the non-targeted columns and index still line up correctly with the columns that are being sampled.

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import pandas as pd
+
+
+def uniform(dataframe, columns, bin_counts):
+    '''
+    Uniformly sample a DataFrame across n-dimensions.
```
jason-curtis, line 6: I haven't figured out quite how to do it, but this function seems like a good application for a [groupby-apply-combine](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html) pattern, which is a super powerful thing in pandas. Something like
`dataframe.groupby(get_bin_from_row).apply(lambda mini_df: mini_df.sample(max_samples)).[concat?]`

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+        [
+            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
+            for i in range(0, len(columns))
+        ],
+    ).transpose()
+    histogram.index = dataframe.index.copy()
+    # Combine all bins into one n-dimensional bin
+    histogram['bin'] = pd.Categorical(histogram.apply(tuple, 1))
+
+    max_samples = histogram['bin'].value_counts().min()
```
jason-curtis: why is this called `max_samples`? I guess it's kind of `max_samples_we_can_get_from_each_bin` but I didn't really get that from the name. `samples_per_bin` would be good since that's what it's used for, and the fact that it comes from the minimum value is clear enough from this line.

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+        Number of bins are assigned positionally to their respective column
+
+    Returns:
+        A pandas DataFrame with a uniform distribution across the specified columns
+    '''
+
+    # Create n-dimensional histogram of requested columns
+    histogram = pd.DataFrame(
+        [
+            pd.cut(dataframe[columns[i]], bin_counts[i], labels=range(0, bin_counts[i]))
```
jason-curtis: do you need to set labels here? Seems like the automatic ranges could be fine.

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,37 @@
...
+        columns: A list of columns names to sample uniformly
+        bin_counts: A list of the number of bins over which to distribute each column
+        Number of bins are assigned positionally to their respective column
+
+    Returns:
+        A pandas DataFrame with a uniform distribution across the specified columns
+    '''
+
+    # Create n-dimensional histogram of requested columns
+    histogram = pd.DataFrame(
```
jason-curtis: as this is implemented, I'm finding `histogram` to be a confusing name. I expected it to be indexed on bins and the values to be the frequency of items in that bin. Maybe this is `binned_dataframe`? or `binned_inputs` or just `binned`

jason-curtis (changes requested):
TFD? Nice.

Mostly stylistic stuff in here, which I think would improve readability.

couple of thoughts on the behavior of this implementation:
1. In working with the actual data set, I found that there were some very small bins on the edge of the data set. When I tried to sample the DataFrame based on that minimum value, I ended up with a very small sample. So my solution was to set the sample size based on the 10th-percentile bin or something. I like your implementation more - it can be a preprocessing step to filter out the smallest bins on the bottom.
2. The way this is implemented, the bin edges are dependent on the data. This means that if we try to "equally sample" two different but similar data sets, we'll end up using similar but different bins. I'm not sure how bad that is. Actually, on second thought: it looks like the user could almost just pass in their own bin bounds if they care about that, and it would get passed through to `pd.cut`. Maybe it'd be best to just document that feature.... or just not worry about it for now.

jason-curtis:
looking forward to using this :)

jaimemarijke:
Just dropping a note here that I glanced through Jason's comments - didn't think too thoroughly about it but it looks like he had some meaty feedback so I'll wait to see how that plays out.

osmo_jupyter/sample.py
EvanSimpson: The `1` is to indicate the columns axis.

EvanSimpson:
I refactored the sampling with your groupby suggestion. I'm not sure it would be simpler or more readable to do a groupby on the input dataframe rather than on the bin data, so I left the bins dataframe, but cleaned it up.

I don't think it should be this function's responsibility to worry about outliers, so I'm going to pass on adding that. 

As for variation in bin bounds, I think for now it would be a bigger problem if we had a dataset that can't saturate our input space in dimesions we care about, than if our code didn't try to fill in for the missing data. That might change if we decide to move to a more real-world setting for data collection, but even then I think we would want to seek out data at the extreme ends.

osmo_jupyter/sample.py
EvanSimpson: This was to cast back from the `Categorical` type to do the value comparison, but with your other suggestions this is unnecessary now.

osmo_jupyter/sample_test.py
jason-curtis: nice, I hadn't realized the part about deliberately unbalancing it here. Much better with the comment :)

osmo_jupyter/sample_test.py
```diff
@@ -0,0 +1,110 @@
...
+            input_dataframe,
+            columns_and_bin_counts={
+                'column_1': 10,
+                'column_2': 10
+            }
+        )
+
+        # Ensure the output rows are still assigned to the correct index value
+        pd.testing.assert_frame_equal(
+            input_dataframe.loc[output_dataframe.index],
```
jason-curtis, line 108: nice!

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # Combine all bins for each row into one series of n-dimensional bins
+    bins = binned.apply(tuple, 1)
+
+    # Take samples from each bin
+    samples_per_bin = bins.value_counts().min()
+    samples_index = bins.groupby(bins).apply(
+        lambda bin_group:
+            bin_group.sample(samples_per_bin)
+    ).index.levels[1].values
```
jason-curtis: minor:

thing I finally figured out yesterday and (mostly at least) understand now: if you `groupby(bins, as_index=False, group_keys=False)` then the group info never gets shoved into the output, so you don't have to do the `.levels[1]` magic number dance. 

`.values` should be redundant here already.

osmo_jupyter/sample_test.py
```diff
@@ -0,0 +1,110 @@
...
+                'one': num_bins
+            },
+        )
+
+        # Use normalized value_counts to get the distribution of values in column 'one'
+        actual_distribution = output_dataframe['one'].value_counts(normalize=True).values
+        # Create an array of expected value count distribution in column 'one'
+        expected_distribution = np.full(num_bins, 1 / num_bins)
+
+        np.testing.assert_almost_equal(actual_distribution, expected_distribution)
```
jason-curtis, line 36: super nitpicky, definitely feel free to ignore:

it feels like you could probably use less code and maybe one slightly magical number for a more straightforward test here. The normalizing and `1/num_bins` complexity kind of cancel each other out - they both make it look like you're doing something sophisticated when really you're just checking that all of the value counts are equal right?

Something like `assert (output_dataframe['one'].value_counts() == 4).all()` might do the trick. If you change how you do it here you'll probably want to do it in `test_sampling_on_multiple_dimensions` too

osmo_jupyter/sample.py
```diff
@@ -0,0 +1,35 @@
...
+    # Create bins for each requested columns
+    binned = pd.DataFrame(
+        {
+            column: pd.cut(dataframe[column], bin_count)
+            for column, bin_count in columns_and_bin_counts.items()
+        },
+    )
+
+    # Combine all bins for each row into one series of n-dimensional bins
+    bins = binned.apply(tuple, 1)
```
jason-curtis: `bins` still doesn't feel like quite the right name to me, but it only exists for like 3 lines so I think it's fine. Could call it `combined_bins`.

Also I'd slightly prefer if you used the kwarg for `axis=1`... maybe it's just a personal thing that the `axis` kwarg tends to trip me up. Keep in mind that `apply` also takes additional args and kwargs and passes them to the function, so using the `axis` kwarg here kind of helps disambiguate.

jason-curtis (approved):
LGTM. thanks for the updates, it feels a lot cleaner!

osmo_jupyter/sample_test.py
EvanSimpson: I think comparing value_counts in `test_sampling_on_multiple_dimensions` would be more difficult. I'm seeing that when the number of columns you are trying to balance is > 2 you don't get a perfect distribution in the output. I haven't quite wrapped my head around why that's true, so I don't know how I would be able to predict the value counts to assert against.

jaimemarijke (commented):
I tried glancing through this PR but decided I was too brain-dead after so many planning meetings to provide good feedback. I'm okay if you want to land without my review, or if you'd like more feedback I will try to take a look tomorrow

PR #27, opened 2019-06-14
Add bin_quantile parameter to uniform sampler

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1125039984561184)

This is an optimization on top of #26, aimed at being less sensitive to outlier data in favor of preserving more data in the output uniform sample.

This adds an optional `bin_quantile` parameter which is used to select the `samples_per_bin` value based on the distribution of the combined bin counts. E.g. `bin_quantile=0.1` will ignore the smallest 10% of bins when determining sample size.

## Manual Testing
See "2019-06-13 Distributions and bins exploration" notebook in [2019-06-13 Distributions and bins exploration](https://drive.google.com/drive/u/0/folders/1m35QHVJaImQKVZgSHL_4OOovK-Cnm2tl) for some example performance against our ml dataset.

### DIFF
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -9,6 +9,8 @@ def uniform(dataframe, columns_and_bin_counts, bin_quantile=0):
         dataframe: A pandas DataFrame
         columns_and_bin_counts: A dictionary of columns names and the number of bins
             over which to distribute each column
+        bin_quantile: The lowest quantile of the combined bin sizes to use to determine
+            the number of samples to take from each bin. Defaults to 0.
 
     Returns:
         A pandas DataFrame with a uniform distribution across the specified columns
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -10,7 +10,7 @@ def uniform(dataframe, columns_and_bin_counts, bin_quantile=0):
         columns_and_bin_counts: A dictionary of columns names and the number of bins
             over which to distribute each column
         bin_quantile: The lowest quantile of the combined bin sizes to use to determine
-            the number of samples to take from each bin. Defaults to 0.
+            the number of samples to take from each bin. Defaults to 0. Empty bins are ignored.
 
     Returns:
         A pandas DataFrame with a uniform distribution across the specified columns
@@ -34,6 +34,8 @@ def uniform(dataframe, columns_and_bin_counts, bin_quantile=0):
         group_keys=False
     ).apply(
         lambda bin_group:
+            # Use min() to ensure sample doesn't throw in the event
+            # samples_per_bin is larger than the number of samples possible
             bin_group.sample(min(samples_per_bin, len(bin_group)))
     ).index
 
--- a/osmo_jupyter/sample.py
+++ b/osmo_jupyter/sample.py
@@ -1,7 +1,7 @@
 import pandas as pd
 
 
-def uniform(dataframe, columns_and_bin_counts):
+def uniform(dataframe, columns_and_bin_counts, bin_quantile=0):
     '''
     Uniformly sample a DataFrame across n-dimensions.
 
@@ -26,13 +26,13 @@ def uniform(dataframe, columns_and_bin_counts):
     combined_bins = binned.apply(tuple, axis=1)
 
     # Take samples from each bin
-    samples_per_bin = combined_bins.value_counts().min()
+    samples_per_bin = combined_bins.value_counts().quantile(bin_quantile).astype(int)
     samples_index = combined_bins.groupby(
         combined_bins,
         group_keys=False
     ).apply(
         lambda bin_group:
-            bin_group.sample(samples_per_bin)
+            bin_group.sample(min(samples_per_bin, len(bin_group)))
     ).index
 
     return dataframe.loc[samples_index]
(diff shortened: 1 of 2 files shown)


osmo_jupyter/sample_test.py
```diff
@@ -108,3 +108,38 @@ def test_sampled_rows_match_input_rows(self):
...
         )
+
+    def test_outliers_ignored_with_quantile(self):
+        num_columns = 2
+        bin_count = 100
+        columns = [f'column_{n}' for n in range(0, num_columns)]
+        input_dataframe = pd.DataFrame(
+            np.vstack((
+                # Generate random array with values between 0-98
+                np.random.randint(0, 98 + 1, (10000, num_columns)),
```
jason-curtis: probably should seed the random number generator for that once-in-a-gazillion event
also this is pretty opaque. why 98+1, while you're using 99 directly below? why 10k?

I'm finding this test pretty hard to follow and I think that rethinking the commenting of this dataframe, and doing a pass to simplify would help. a couple of things I can put my finger on:
* It took me a while to figure out that 99 is going to be the outlier bin and the `randint` is supposed to ensure that there's more than 99 values in every other bin. I had to do mental math and stuff.
* the use of `num_columns` as a variable made it feel like it was generalizable. IMO, the test would be simpler and easier to grok if it was just straight-up `columns = ['column_for_binning', 'other_column']`. That said, since you're not asserting anything about the other column, do you need to have it all? It seems like it's what led you to take on the complexity of the `vstack` which is another piece of cognitive load we could do without. If you want a second column you could just set it to a constant.

osmo_jupyter/sample_test.py
```diff
@@ -108,3 +108,38 @@ def test_sampled_rows_match_input_rows(self):
...
+
+        unadjusted_samples = module.uniform(
+            input_dataframe,
+            columns_and_bin_counts={
+                'column_1': bin_count
+            }
+        )
+
+        # Expect the outlier to force max sample size to 1 sample per bin
+        assert len(unadjusted_samples) == bin_count
```
jason-curtis: this could be a little more self-documenting; there's implicit math happening here that could be explicit. something like
```python
actual_samples_per_bin = len(unadjusted_samples) / bin_count
assert actual_samples_per_bin == n_outlier_samples
```

osmo_jupyter/sample_test.py
```diff
@@ -108,3 +108,38 @@ def test_sampled_rows_match_input_rows(self):
...
+        outlier_adjusted_samples = module.uniform(
+            input_dataframe,
+            columns_and_bin_counts={
+                'column_1': bin_count
+            },
+            bin_quantile=0.1
+        )
+
+        # Expect the bin_quantile param to ignore the outlier data
+        assert len(outlier_adjusted_samples) > bin_count
```
jason-curtis: I'm on the fence about asking for this to be two separate tests with a helper function. It could come out nice and readable, and have the advantage of not shadowing. I see why they logically go as one unit but they could also stand alone just fine. I guess I'd just like to hear that you considered it :)

Also, either way, I think you could stick to the setup/call/assert pattern here and leave both asserts to the end - I think that would make the intent between the two sub-tests quicker to grok.

osmo_jupyter/sample.py
```diff
@@ -1,14 +1,16 @@
...
+def uniform(dataframe, columns_and_bin_counts, bin_quantile=0):
     '''
     Uniformly sample a DataFrame across n-dimensions.
 
     Args:
         dataframe: A pandas DataFrame
         columns_and_bin_counts: A dictionary of columns names and the number of bins
             over which to distribute each column
+        bin_quantile: The lowest quantile of the combined bin sizes to use to determine
+            the number of samples to take from each bin. Defaults to 0.
```
jason-curtis: we may want to update this at some point to a standard value, but maybe that'll happen as part of [design spike: define the target](https://app.asana.com/0/819671808102776/1126912711025459/f)

also, is it worth calling out that empty bins will be ignored no matter what, or is that obvious?

osmo_jupyter/sample.py
```diff
@@ -26,13 +28,13 @@ def uniform(dataframe, columns_and_bin_counts):
...
     # Take samples from each bin
-    samples_per_bin = combined_bins.value_counts().min()
+    samples_per_bin = combined_bins.value_counts().quantile(bin_quantile).astype(int)
     samples_index = combined_bins.groupby(
         combined_bins,
         group_keys=False
     ).apply(
         lambda bin_group:
-            bin_group.sample(samples_per_bin)
+            bin_group.sample(min(samples_per_bin, len(bin_group)))
```
jason-curtis, line 39: this probably could use a comment about how `min` prevents this blowing up

jason-curtis (approved):
I'd like to see some simplification of the test, but don't necessarily need to do another review afterwards unless you want one.

jaimemarijke:
Dang, Jason beat me to the punch again. I agree with his comments - I'll plan to review any changes.

osmo_jupyter/sample_test.py
EvanSimpson: I ended up going with 2 tests for readability. Having a helper function set up the dataset makes it much simpler, logically.

jaimemarijke (approved):
Nice tests! It was easy for me to understand why they existed and how they worked.

PR #28, opened 2019-06-17
DO -> mmHg, plus a couple of related notebook features

jason-curtis:
[Asana Ticket: Use partial pressure in DO fit libraries](https://app.asana.com/0/819671808102776/1123633285348210/f)

## Steps taken

Since this is a search-and-destroy ticket, I tried to keep a pretty detailed log so you can see if you think of anything I missed:

- search for "%" in osmo-jupyter, check that all are appropriate and convert % sat to mmHg
- search for "sat" in osmo-jupyter, do the same
- update calibration data prepper to generate partial pressure and not round temperature (the latter was a procedure change in the notebook fork since we trust the YSI enough to want to see its temperature data)
- update WORKING_DO_PARAMS by multiplying the appropriate ones (A_k_sv1 and A_k_sv2) by 100/160 (ratio of 100% sat to the sometimes-equivalent partial pressure). no change to test fixture values was necessary for this. Yay math
- Update constants to derive everything from as few source constants as possible -> very minor change to some test fixture values due to DO_PARTIAL_PRESSURE_MMHG_AT_1ATM changing from 160 to ~159.2
- port error bounds lines update from 2019-05-17 All Passive Calibration Sweeps Organization - now `calibration_error_plot()` includes those red lines at our DO error bounds.
- Tested that the [latest calibration notebook](https://drive.google.com/open?id=1N6NQ6-7R_Opi4_1AP3MY-7xL3UogVnWp) still runs after deleting the forked code from it. (Note: I didn't convert the data-prep part of that notebook - it uses multiple YSI and camera data files, which feels too specialized to librarify)

## Manual Testing

Deleted ~200 lines of forked code from the notebook linked above, and imported new stuff instead. Restart and run all. The plots still work though the calibration still looks pretty bad:
image

## TODO

before landing, run the docs generator


osmo_jupyter/calibration/do/prep.py
```diff
@@ -1,8 +1,14 @@
 import pandas as pd
 
+from osmo_jupyter.constants import FRACTION_O2_IN_ATMOSPHERE
 from osmo_jupyter.ysi import join_interpolated_ysi_data
 
 
+def _get_o2_partial_pressure(atmospheric_pressure_mmhg, percent_saturation):
+    saturation_fraction = percent_saturation * 0.01
```
jaimemarijke: nit: maybe consistency with saturation being a postfix or prefix? s/saturation_fraction/fraction_saturation

jaimemarijke: other nit: divide by 100 instead of multiply by 0.01 would save me about 0.01 mental frictions

osmo_jupyter/calibration/do/prep_test.py
```diff
@@ -12,14 +12,16 @@ def ysi_data_filepath(tmpdir):
     pd.DataFrame([
         {
             'Timestamp': datetime(2019, 1, 1, 1, 1, 45),
-            'Dissolved Oxygen (%)': 1,
+            'Dissolved Oxygen (%)': 100,
```
jaimemarijke, line 15: I'm confused: why did this go from 1 to 100?

osmo_jupyter/constants.py
```diff
@@ -5,22 +5,33 @@
 
 DEGREES_CELSIUS_AT_ZERO_KELVIN = 273.15
 
-# Full range of possible (in normal circumstances) DOs, in percent sat
-DO_MIN = 0
-DO_MAX = 100
+# https://en.wikipedia.org/wiki/Atmosphere_(unit)
+ONE_ATM_MMHG = 760.001
```
jaimemarijke: I call BS on the 0.001. Isn't 760 just sort of an average anyway?

osmo_jupyter/constants.py
```diff
@@ -5,22 +5,33 @@
...
+
+# https://www.wolframalpha.com/input/?i=fraction+of+oxygen+in+atmosphere
+FRACTION_O2_IN_ATMOSPHERE = 0.20948
 
 # Corresponds to 100% saturation at 1atm - use to convert between partial pressure and saturation
-DO_PARTIAL_PRESSURE_MMHG_AT_1ATM = 160
+DO_PARTIAL_PRESSURE_MMHG_AT_1ATM = ONE_ATM_MMHG * FRACTION_O2_IN_ATMOSPHERE
+
+# Full range of possible (in normal circumstances) DO partial pressures, in mmHg
```
jaimemarijke, line 17: Does a slightly high pressure day count as a normal circumstance?

osmo_jupyter/constants.py
```diff
@@ -5,22 +5,33 @@
...
+DO_MAX_MMHG = DO_PARTIAL_PRESSURE_MMHG_AT_1ATM
 
 # Assume sea level, no salt, standard temp;
 # details of how much this matters are discussed in 2019-02-18 updated DN error bounds
 DO_SATURATION_AT_SEA_LEVEL_MG_PER_L = 8.26
+STANDARD_MG_PER_L_PER_MMHG = DO_SATURATION_AT_SEA_LEVEL_MG_PER_L / DO_PARTIAL_PRESSURE_MMHG_AT_1ATM
+
+# Error bounds
+ERROR_BOUNDS_MG_PER_L = 0.3
+ERROR_BOUNDS_MMHG = ERROR_BOUNDS_MG_PER_L / STANDARD_MG_PER_L_PER_MMHG
```
jaimemarijke, line 28: I know 0.3 mg/L has been our standard error bound forever, but.. I wish it was defined in terms of mmHg first.

osmo_jupyter/simulate/do_patch_test.py
```diff
@@ -10,7 +10,7 @@ class TestNormalizedReading:
         'name, do, temperature, expected_normalized_reading',
         [
             ('max fluorescence at low DO, low temp', 0, TEMPERATURE_STANDARD_OPERATING_MIN, 1),
```
jaimemarijke: since you replaced the 100's with `DO_MAX_MMHG`, should you also replace the 0's with `DO_MIN_MMHG`?

jaimemarijke (approved):
I don't think my review would catch any major errors, since it's pretty hard to catch those just by reading the code.

osmo_jupyter/calibration/do/prep_test.py
jason-curtis: I just changed it to make the expected DO values a little more intuitive.

osmo_jupyter/constants.py
jason-curtis: hmm do you think we should just multiply it by 1.1 or something?

osmo_jupyter/constants.py
jason-curtis: I disagree? maybe? I'm not sure if (a)you actually want me to change this (disagree), or (b) if this is just a gripe about how the industry might be somewhat broken (agree).

I think it's better to state the user-space requirement and then derive other quantities from it,  as I did here.

osmo_jupyter/constants.py
jason-curtis: found a less precise source :+1:

osmo_jupyter/simulate/do_patch_test.py
```diff
@@ -21,8 +22,8 @@ def test_normalized_reading_bounds(self, name, do, temperature, expected_normali
         [
             ('middlin DO at min temp', 50, TEMPERATURE_STANDARD_OPERATING_MIN),
```
EvanSimpson: Is 50 still an appropriate mid value here?

osmo_jupyter/calibration/do/prep.py
```diff
@@ -1,8 +1,14 @@
 import pandas as pd
 
+from osmo_jupyter.constants import FRACTION_O2_IN_ATMOSPHERE
 from osmo_jupyter.ysi import join_interpolated_ysi_data
 
 
+def _get_o2_partial_pressure(atmospheric_pressure_mmhg, percent_saturation):
+    fraction_saturation = percent_saturation / 100
+    return fraction_saturation * FRACTION_O2_IN_ATMOSPHERE * atmospheric_pressure_mmhg
```
EvanSimpson: It took me a second to wrap my head around why we're doing `fraction_ * fraction_`.  I'm not sure if there's a better name for `fraction_saturation` but it wasn't immediately clear that `FRACTION_O2_IN_ATMOSPHERE ` represents 100 percent sat and we're taking the fraction of that.

EvanSimpson (approved):
I just had a couple of small notes.

osmo_jupyter/calibration/do/prep.py
jason-curtis: tweaking it a bit. I think it'll read more clearly and have the consistency @jaimemarijke was looking for (at least internally) if we use `saturation_percent` and `saturation_fraction` here rather than `percent_saturation` and `fraction_saturation`.

osmo_jupyter/simulate/do_patch_test.py
jason-curtis: hmm it's still kinda middlin but I'll up it to the actual middle because why not

PR #29, opened 2019-06-17
Quick JFFI Prototype: Automatic changes from running `black`

jason-curtis:
No asana ticket.

A while ago I came across [Black: the uncompromising code formatter](https://github.com/python/black). Finally got around to trying it.

## Overall impressions:

* `black` deals with a lot of formatting stuff that I'd like to have done, but don't want to do. Splitting imports into proper ~alphabetized~ (edit: not alphabetized) parenthesized instead of backslash'd sequences; proper spacing around operators, proper indentation on nested things.
* `black` is very fast. Much faster than `flake8 --fix`
* setup instructions were straightforward and correct for integrating with PyCharm and other editors are supported too. There's even a jupyter magic command we could use to auto-format jupyter code :open_mouth: though it's a bit hacky.
* `black` actually fixed stuff without breaking anything other than a few flake things for which they explain in their docs why flake8 is wrong. I also had to move one flake-ignore comment that `black` moved to the wrong line when it reformatted stuff. But in general, they claim that it leaves the AST identical to how it was beforehand, so it can't introduce bugs.

## Differences between our style and Black:

* double quotes, not single. Their recommendation is that if you want to save keystrokes, you should just type in single quotes and then let `black` fix it for you. I'd go for that.
* line length: Black will make code lines shorter if they're longer than 88 characters. However, it doesn't know how to shorten strings that are too long. That seems fine with me. If we leave flake8 configured as-is, we'll only get flake errors when things exceed 120 but Black will helpfully try and keep things to a compact 88. We could also tell it to use 120 but that's a little counter to black's philosophy of "whatever color you want [so long as it is black]".
* sometimes we nest multiple levels/types of parens on one line where Black will expand them. e.g.
```
        array = np.array([
            somethingorother
        ])
```
becomes
```
        array = np.array(
            [
                somethingorother
            ]
        )
```
which I actually quite like
* Sometimes we use line breaks in math or in arrays to give the code more of an intuitively correct shape, eg. we might do
```
        even_array = np.array([
            [1, 1, 2, 2],
            [1, 1, 2, 2],
            [3, 3, 4, 4],
            [3, 3, 4, 4],
        ])
```
but `black` will turn that into 
```
even_array = np.array([[1, 1, 2, 2], [1, 1, 2, 2], [3, 3, 4, 4], [3, 3, 4, 4]])
```
as long as it fits within the line length. I'm on the fence of whether that's acceptable. It does feel like it would be at least an initial hit in readability.

## Manual Testing

* ran `tox -r`

## Suggestion

* try this out on this repo:
    * land the branch (includes auto-checking as part of tests)
    * set `black` up in our editors or run it manually before commit
    * write up a ticket to implement `black` in other repos if we like it in a week or two

What do y'all think?


EvanSimpson:
Not being able to format readable arrays is unfortunate, but I am willing to trade having opinions about whitespace for not having to think about whitespace.

I would anticipate a lot of 
"Why did you choose to format it this way?" 
"I didn't, `black` did"
"hrm"
in code review for a short time though.

EvanSimpson (approved):

osmo_jupyter/calibration/do/curve_test.py
```diff
@@ -45,8 +46,6 @@ class TestGetArrheniusRate:
     def test_get_arrhenius_rate(self):
         # Spot-check rate function with some actual numbers
         actual = module._get_arrhenius_rate(
-            temperature_c=0,
-            preexponential_factor=10,
-            activation_energy=1e-4
+            temperature_c=0, preexponential_factor=10, activation_energy=1e-4
```
jaimemarijke, line 49: yuck. I really don't like this

jaimemarijke (approved):
I'm torn on this. I was leaning slightly against it because these two things really grate on me:

```
# in:

even_array = np.array([
    [1, 1, 2, 2],
    [1, 1, 2, 2],
    [3, 3, 4, 4],
    [3, 3, 4, 4],
])

# out: 

even_array = np.array([[1, 1, 2, 2], [1, 1, 2, 2], [3, 3, 4, 4], [3, 3, 4, 4]])
```

```
# in:

ImportantClass.important_method(exc, limit, lookup_lines, capture_locals, extra_argument)

# out:

ImportantClass.important_method(
    exc, limit, lookup_lines, capture_locals, extra_argument
)
```

However, I poked into a bit and discovered that you can use `# fmt: off` and `# fmt: on`  to disable/re-enable black for blocks of code. 

With that ability to save us from any particularly aggravating re-formatting, I say ship it! 
Let's make sure to:

- add it to all our repos
- update code style manifesto
- update our personal IDEs

jason-curtis:
cool, I filed https://app.asana.com/0/819671808102776/1127643074336798/f

PR #31, opened 2019-06-26
JFFI: Consolidate configuration into tox.ini.

jason-curtis:
pytest and flake8 already look there, so why not?

Plus there is an issue with atom in particular where for some reason, flake8 does not look in setup.cfg for configuration.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ commands =
 
 [flake8]
 exclude = .git,__pycache__,.eggs,.pytest_cache,build,dist,lib
-max_line_length = 10
+max_line_length = 120
 ignore = W503,E203
 
 [pytest]
--- a/tox.ini
+++ b/tox.ini
@@ -11,3 +11,11 @@ deps =
     -rtest_requirements.txt
 commands =
     pytest []
+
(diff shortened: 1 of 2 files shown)


EvanSimpson (approved):

jaimemarijke (approved):
👍 to fewer config files.

tox.ini
```diff
@@ -11,3 +11,11 @@ deps =
     -rtest_requirements.txt
 commands =
     pytest []
+
+[flake8]
+exclude = .git,__pycache__,.eggs,.pytest_cache,build,dist,lib
+max_line_length = 120
+ignore = W503,E203
```
rush340, line 18: why E203?

jaimemarijke: good question. I don't remember any particular conversation about it. Looking at the rule, I would prefer to have the rule enforced, not ignored. https://lintlyci.github.io/Flake8Rules/rules/E203.html

jason-curtis: because it's not pep8 and black has reason to violate it sometimes.

https://github.com/python/black#slices

rush340 (approved):

PR #32, opened 2019-08-06
JFFI: upgrade plotly to v4

jason-curtis:
No asana.

### Why I think this is a good idea

plotly V4 provides direct access to the plotly_express package which is super convenient. It also makes several other nice changes.

### Why I think it's safe

I have run a few recent notebooks with plotly v4, and skimmed the [migration guide](https://plot.ly/python/v4-migration/) and I don't think any of the breaking changes will affect us in our recent notebooks. If they do (most likely we're making now-unnecessary `plotly.offline.init_notebook_mode()` calls in some old notebooks) then it'll be trivial to update notebooks when we update the notebook's version of osmo-jupyter.

### What changes to expect

mostly a new default theme and access to `plotly_express` without installing a separate package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         "numpy",
         "pandas",
         "petl",
-        "plotly >= 3, < 4",
+        "plotly >= 4, < 5",
         "pymysql",
         "requests",
         "scipy",


EvanSimpson (approved):
Should we go through all repos that have plotly as a dependency and update them?

jason-curtis:
yeah, `calibration-environment` is already pinned to `>=4` (which maybe should be more strict...) but if this passes I'll go ahead and update [cosmobot-process-experiment](https://github.com/OsmoSystems/cosmobot-process-experiment/blob/master/setup.py) too.

jaimemarijke (approved):
Is there a v5 out yet? Seems a little simpler to just drop the version requirements unless we have a problem

jason-curtis:
> Is there a v5 out yet? Seems a little simpler to just drop the version requirements unless we have a problem

we started doing this because we did have some breaking changes with upgrading to v3. plotly continues to actively release major versions, and I wouldn't be surprised if v5 had breaking changes we cared about. I think we should stick with this because otherwise future notebooks will be subject to those changes.

that said, I'm not a fan of manually managing versions and if you or anyone feels strongly I'll happily be convinced to set this to 4+

PR #33, opened 2019-08-15
Add tools for processing calibration data

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1135264514117500)

Adds some helper tools for processing all of our various sources of data into neat datasets.
Tests feel pretty messy and coverage a little light due to the amount of mock data necessary to adequately test a lot of this functionality. 

Also open to feedback re: spitting up / renaming the module.

## Manual Testing
See some example usage on our most recent collection of data in `Move some dataset generation code into osmo-jupyter.ipynb` [here](https://drive.google.com/drive/u/0/folders/1aZ3WLY2amdifzrmaQIp_5pRbanRh2N2K)

### DIFF
--- a/osmo_jupyter/dataset.py
+++ b/osmo_jupyter/dataset.py
@@ -1,0 +1,202 @@
+import os
+from typing import List
+
+import numpy as np
+import pandas as pd
+
+
+def _interpolate_and_join_data(df_1: pd.DataFrame, df_2: pd.DataFrame) -> pd.DataFrame:
+    # Upsample df_1 to every second to allow any timestamp to be joined on a strict match
+    resampled_df_1 = df_1.resample("s").interpolate(method="slinear")
+
+    return df_2.join(  # By default join happens on index values
+        resampled_df_1, how="inner"  # Drop any rows without a match in both DataFrames
+    )
+
+
+def open_picolog_file(picolog_filepath):
+    return (
+        pd.read_csv(
+            picolog_filepath,
+            index_col=0,  # first column, is unlabeled
+            parse_dates=[0],  # first column
+            date_parser=lambda col: pd.to_datetime(col, utc=False).tz_localize(
+                None
+            ),  # Remove timezone information
+        )
+        .rename_axis("timestamp")
+        .add_prefix("PicoLog ")
+    )
+
+
+def open_calibration_log_file(calibration_log_filepath):
+    return pd.read_csv(
+        calibration_log_filepath,
+        index_col="timestamp",
+        parse_dates=["timestamp"],
+        date_parser=lambda col: pd.to_datetime(col, utc=False).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        ),  # Truncate fractional seconds
+    )
+
+
+def open_and_combine_picolog_and_calibration_data(
+    calibration_log_filepaths: List[str], picolog_log_filepaths: List[str]
+) -> pd.DataFrame:
+    picolog_data = pd.concat(
+        [
+            open_picolog_file(picolog_filepath)
+            for picolog_filepath in picolog_log_filepaths
+        ],
+        sort=True,
+    )
+
+    calibration_data = pd.concat(
+        [
+            open_calibration_log_file(calibration_log_filepath)
+            for calibration_log_filepath in calibration_log_filepaths
+        ],
+        sort=True,
+    )
+
+    return _interpolate_and_join_data(picolog_data, calibration_data)
+
+
+def get_equilibration_boundaries(equilibration_status: pd.Series) -> pd.DataFrame:
+    # Get a boolean series with timestamped index of equilbrated data points
+    equilibrated_mask = equilibration_status == "equilibrated"
+
+    # Upsample to every second
+    equilibration_status_sampler = equilibrated_mask.astype(int).resample("s")
+
+    # In order to not interpolate any gaps around equilibrated states as also equilibrated, interpolate
+    # state as an integer (True == 1) and truncate any decimals (e.g. 0.2 -> 0, or False)
+    interpolated_equilbration_status = equilibration_status_sampler.interpolate().apply(
+        np.floor
+    )
+
+    # rolling difference from previous row
+    leading_edges = interpolated_equilbration_status[
+        interpolated_equilbration_status.astype(int).diff() == 1
+    ].index
+    # rolling difference with next row
+    trailing_edges = interpolated_equilbration_status[
+        interpolated_equilbration_status.astype(int).diff(periods=-1) == 1
+    ].index
+
+    # Trim edges to align
+    if len(leading_edges) > len(trailing_edges):
+        leading_edges = leading_edges[:-1]
+    if len(trailing_edges) > len(leading_edges):
+        trailing_edges = trailing_edges[1:]
+
+    return pd.DataFrame(
+        {"leading_edge": leading_edges, "trailing_edge": trailing_edges}
+    )
+
+
+def pivot_process_experiment_results_on_ROI(
+    experiment_df: pd.DataFrame,
+    ROI_names: List[str] = None,
+    msorm_types: List[str] = ["r_msorm", "g_msorm", "b_msorm"],
+) -> pd.DataFrame:
+    """
+    Uses a pivot table to flatten multiple ROIs for the same image into one row.
+    Returns a DataFrame with one row per image, and msorm values for a subset of ROIs.
+    """
+    if ROI_names:
+        experiment_df = experiment_df[experiment_df["ROI"].isin(ROI_names)]
+
+    # Pivot creates a heirachical multiindex data frame with an index for each 'values' column
+    pivot = experiment_df.pivot(
+        index="timestamp", columns="ROI", values=msorm_types + ["image"]
+    )
+
+    # There's one copy of the image name for each ROI due to the pivot, so pull out the top level index
+    images = pivot["image"][experiment_df["ROI"].values[0]]
+    pivot.drop("image", axis=1, inplace=True)
+    # Flatten the pivot table index, and add images back in
+    pivot.columns = [" ".join(col[::-1]).strip() for col in pivot.columns.values]
+    pivot["image"] = images
+
+    return pivot
+
+
+def open_and_combine_process_experiment_results(
+    results_filepaths: List[str],
+    ROI_names: List[str] = None,
+    msorm_types: List[str] = ["r_msorm", "g_msorm", "b_msorm"],
+) -> pd.DataFrame:
+    all_roi_data = pd.concat(
+        [
+            pd.read_csv(results_filepath, parse_dates=["timestamp"])
+            for results_filepath in results_filepaths
+        ]
+    )
+
+    return pivot_process_experiment_results_on_ROI(all_roi_data, ROI_names, msorm_types)
+
+
+def get_all_experiment_images(
+    local_sync_directory: str, experiment_names: List[str]
+) -> pd.DataFrame:
+    all_images = pd.concat(
+        [
+            pd.DataFrame(
+                {
+                    "experiment": experiment_name,
+                    "image": os.listdir(
+                        os.path.join(local_sync_directory, experiment_name)
+                    ),
+                },
+                dtype="object",  # Ensure correct dtype when no images are found
+            )
+            for experiment_name in experiment_names
+        ]
+    )
+
+    # Filter out experiment log files
+    return all_images.drop(all_images[~all_images["image"].str.contains("jpeg")].index)
+
+
+def filter_equilibrated_images(equilibration_range, image_ROI_data):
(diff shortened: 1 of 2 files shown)


osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,202 @@
...
+    if ROI_names:
+        experiment_df = experiment_df[experiment_df["ROI"].isin(ROI_names)]
+
+    # Pivot creates a heirachical multiindex data frame with an index for each 'values' column
+    pivot = experiment_df.pivot(
+        index="timestamp", columns="ROI", values=msorm_types + ["image"]
+    )
+
+    # There's one copy of the image name for each ROI due to the pivot, so pull out the top level index
+    images = pivot["image"][experiment_df["ROI"].values[0]]
```
EvanSimpson, line 161: I looked into instead doing this transform with a `groupby` and `apply`ing the pivot to clean up this messy bit at the end, but the performance was significantly worse than just pivoting everything altogether.

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,202 @@
...
+            open_calibration_log_file(calibration_log_filepath)
+            for calibration_log_filepath in calibration_log_filepaths
+        ],
+        sort=True,
+    )
+
+    return _interpolate_and_join_data(picolog_data, calibration_data)
+
+
+def get_equilibration_boundaries(equilibration_status: pd.Series) -> pd.DataFrame:
```
jason-curtis, line 86: it would be good to have a docstring for this, it wasn't initially clear why using this was better than just joining first and then filtering. something like 
``` Get a DataFrame of equilibration periods with their leading and trailing edges
Takes a conservative approach, assuming all time between unequilibrated and equilibrated input points are unequilibrated.
```

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,202 @@
...
+    # Get a boolean series with timestamped index of equilbrated data points
+    equilibrated_mask = equilibration_status == "equilibrated"
+
+    # Upsample to every second
+    equilibration_status_sampler = equilibrated_mask.astype(int).resample("s")
+
+    # In order to not interpolate any gaps around equilibrated states as also equilibrated, interpolate
+    # state as an integer (True == 1) and truncate any decimals (e.g. 0.2 -> 0, or False)
+    interpolated_equilbration_status = equilibration_status_sampler.interpolate().apply(
```
jason-curtis: it's not clear to me that we need to interpolate. it feels like the end result of the truncation scheme here is that the timestamps in the output where the equilibration status changes, are the same as the timestamps on the inputs where the status changed:

```
original: 0      0     1     1    0
interp'd: 0000000000000111111100000
diff()  : 0000000000000+000000-0000
leading:               *
diff(-1): 0000000000001000000+00000
trailing:                    *
```
so if you skipped this step and then just used `equilibrated_mask` on lines 79 and onwards, wouldn't you get just about the same thing?

Also we may want to consider if there are repercussions here for how we should structure the calibration runner. There may be edge cases in which we're telling the pi to take pictures but which won't count as equilibrated by this metric. Which brings up the question - do we expect to actually use this function once Michael's current work lands?

osmo_jupyter/dataset_test.py
```diff
@@ -0,0 +1,323 @@
...
+class TestFilterEquilibratedImages:
+    def test_returns_only_equilibrated_images(self, mock_calibration_file_obj):
+        test_roi_data = pd.DataFrame(
+            [
+                {"timestamp": pd.to_datetime("2019-01-01"), "image": "image-0.jpeg"},
+                {"timestamp": pd.to_datetime("2019-01-03"), "image": "image-1.jpeg"},
+            ]
+        ).set_index("timestamp")
+
+        test_equilibration_boundaries = pd.DataFrame(
```
jason-curtis: nit: why'd you construct a `DataFrame` here with one row and then use `.iloc`? Why not just make a `Series` in the first place?

jason-curtis (changes requested):
awaiting docstrings

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+    leading_edges = equilibrated_mask[equilibrated_mask.astype(int).diff() == 1].index
+    trailing_edges = equilibrated_mask[
+        equilibrated_mask.astype(int).diff(periods=-1) == 1
+    ].index
+
+    # Trim edges to align
+    if len(leading_edges) > len(trailing_edges):
+        leading_edges = leading_edges[:-1]
+    if len(trailing_edges) > len(leading_edges):
+        trailing_edges = trailing_edges[1:]
```
jason-curtis: minor: this stuff feels like it wants a parameterized test case

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+def open_and_combine_process_experiment_results(
+    process_experiment_result_filepaths: List[str],
+    ROI_names: List[str] = None,
+    msorm_types: List[str] = ["r_msorm", "g_msorm", "b_msorm"],
+) -> pd.DataFrame:
+    """
+        Open multiple process experiment result files and combine into a single DataFrame with one row per image.
+
+        Args:
+            process_experiment_result_filepaths: A list of filepaths to process experiment summary statistics files.
```
jason-curtis, line 187: Not sure what'll happen if ROIs don't match; do you know? I'm fine leaving it explicitly undefined
```suggestion
            process_experiment_result_filepaths: A list of filepaths to process experiment summary statistics files.
                These are all expected to have the same ROIs.
```

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+    # Pivot creates a heirachical multiindex data frame with an index for each 'values' column
+    pivot = experiment_df.pivot(
+        index="timestamp", columns="ROI", values=msorm_types + ["image"]
+    )
+
+    # There's one copy of the image name for each ROI due to the pivot, so pull out the top level index
+    images = pivot["image"][experiment_df["ROI"].values[0]]
+    pivot.drop("image", axis=1, inplace=True)
+    # Flatten the pivot table index, and add images back in
+    pivot.columns = [" ".join(col[::-1]).strip() for col in pivot.columns.values]
```
jason-curtis: minor: logic here is a little surprising. not sure if black will let you sneak a comment in there but it'd be nice
```suggestion
    pivot.columns = [
       # ("r_msorm", "some ROI name") -> "some ROI name r_msorm"
        " ".join(col[::-1]).strip()
        for col in pivot.columns.values
    ]
```

osmo_jupyter/dataset.py
jason-curtis: minor: I'm definitely struggling to picture the DF in each stage of this transformation. maybe a couple more comments would help (right now my sticking point is imagining exactly what the indices are after the pivot and how those end up as columns on line 140)? or not; I think the unit test and docstring convince me this is doing the right thing.

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+            pd.DataFrame(
+                {
+                    "experiment": experiment_name,
+                    "image": os.listdir(
+                        os.path.join(local_sync_directory, experiment_name)
+                    ),
+                },
+                dtype="object",  # Ensure correct dtype when no images are found
+            )
+            for experiment_name in experiment_names
```
jason-curtis, line 221: nit: it might be slightly more straightforward to just construct this DF straight-up with a nested list comprehension... but that option invokes less broadcasting magic :)
`pd.DataFrame([{...} for experiment in experiments for image in os.listdir(...)`

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Args:
+            equilibration_range: A Series with indexed start_time and end_time values
+            df: A datetime indexed DataFrame
+        Returns:
+            A fitlered df containing only values whose index is between the given start_time and end_tim.
+    """
+    leading_edge_mask = df.index > equilibration_range["start_time"]
+    trailing_edge_mask = df.index < equilibration_range["end_time"]
+    return df[leading_edge_mask & trailing_edge_mask]
```
jason-curtis, line 245: I :heart: pandas/numpy

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+
+def filter_equilibrated_images(equilibration_range: pd.Series, df: pd.DataFrame):
+    """
+        Filter a datetime indexed DataFrame to a given equilibration range
+
+        Args:
+            equilibration_range: A Series with indexed start_time and end_time values
+            df: A datetime indexed DataFrame
+        Returns:
+            A fitlered df containing only values whose index is between the given start_time and end_tim.
```
jason-curtis: nit. typing a little fast now, are we
```suggestion
            A filtered df containing only values whose index is between the given start_time and end_time.
```

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+    all_images = pd.concat(
+        [
+            pd.DataFrame(
+                {
+                    "experiment": experiment_name,
+                    "image": os.listdir(
+                        os.path.join(local_sync_directory, experiment_name)
+                    ),
+                },
+                dtype="object",  # Ensure correct dtype when no images are found
```
jason-curtis: minor: looks like there's another edge case here that could use a regression test

jason-curtis (approved):
nothin' major! I only did a fairly cursory look at the tests.

osmo_jupyter/dataset.py
EvanSimpson: Oh, I think it will just produce a bunch on NaN values. I can verify, but I'll add a comment about that.

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,288 @@
...
+    # timestamp    image     DO Patch r_msorm   Reference Patch r_msorm...
+    # 2019-01-01       1                   .4                        .5
+
+    return pivot
+
+
+def open_and_combine_process_experiment_results(
+    process_experiment_result_filepaths: List[str],
+    ROI_names: List[str] = None,
+    msorm_types: List[str] = ["r_msorm", "g_msorm", "b_msorm"],
```
rush340: minor: mutable defaults make me nervous (since all callers get the same instance). Could set to None and assign in the function, or use a tuple for the default.

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def open_and_combine_and_filter_source_data(
+    local_sync_directory: str,
+    experiment_names: List[str],
+    calibration_log_filepaths: List[str],
+    picolog_log_filepaths: List[str],
+    process_experiment_result_filepaths: List[str],
+    ROI_names: List[str] = None,
+    msorm_types: List[str] = ["r_msorm", "g_msorm", "b_msorm"],
```
rush340: same as above

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,288 @@
...
+    # image     ROI             r_msorm     timestamp
+    #     1     DO patch             .4     2019-01-01
+    #     1     Reference Patch      .5     2019-01-01
+
+    # Pivot creates a heirachical multiindex data frame with an index for each 'values' column
+    pivot = experiment_df.pivot(
+        index="timestamp", columns="ROI", values=msorm_types + ["image"]
+    )
+
+    # After pivot -> Multi-level index
```
rush340, line 155: These comments are very helpful!

osmo_jupyter/dataset.py
```diff
@@ -0,0 +1,288 @@
...
+
+    if (
+        len(trailing_edges)
+        and len(leading_edges)
+        and trailing_edges[0] < leading_edges[0]
+    ):
+        trailing_edges = trailing_edges[1:]
+
+    if len(trailing_edges) > len(leading_edges):
+        trailing_edges = trailing_edges[1:]
```
rush340: Could we instead add the first item to leading edges (or last item to trailing edges in the "if" below this) so that we don't lose equilibrated ranges at the beginning/end of a dataset?

rush340 (approved):
Looks good!

osmo_jupyter/dataset.py
EvanSimpson: I think duplicating a final leading edge to also be a trailing edge makes sense, since that could come up if the calibration software crashed in an equilibrated state, but I'm hesitant about keeping data that starts out equilibrated. In theory that shouldn't ever happen and would mean the data had been corrupted, since you have to have 5 minutes of samples before anything is labeled as equilibrated, right?

jason-curtis: this makes sense to me, but it's subtle enough that it'd be nice to at least have a comment. I think I might also be leaning slightly towards the symmetrical approach (include extra leading edges too) since it's conceivable that we could end up using this function with a sub-section of a dataset.

PR #34, opened 2019-08-27
Bugfix: Include sensor data in dataset magic function

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1137324533382466)

Joins picolog and calibration environment data back on to image ROI data in `open_and_combine_and_filter_source_data`. Previously it was only using the calibration data to filter out unequilibrated data.

## Manual Testing
Fiddled with this in a notebook, but updated the test to check for this.

### DIFF
--- a/osmo_jupyter/dataset_test.py
+++ b/osmo_jupyter/dataset_test.py
@@ -408,6 +408,7 @@ class TestOpenAndCombineSourceData:
                         "ROI 0 g_msorm": 0.6,
                         "ROI 1 g_msorm": 0.3,
                         "image": "image-1.jpeg",
+                        "PicoLog Temperature Ave. (C)": 39.75,
                         "experiment": experiment_name,
                     }
                 ]
--- a/osmo_jupyter/dataset_test.py
+++ b/osmo_jupyter/dataset_test.py
@@ -15,10 +15,26 @@ TEST_PICOLOG_DATA = pd.DataFrame(
 
 TEST_CALIBRATION_DATA = pd.DataFrame(
     [
-        {"timestamp": "2019-01-01 00:00:00.1", "equilibration status": "waiting"},
-        {"timestamp": "2019-01-01 00:00:01.1", "equilibration status": "equilibrated"},
-        {"timestamp": "2019-01-01 00:00:03.1", "equilibration status": "equilibrated"},
-        {"timestamp": "2019-01-01 00:00:04.1", "equilibration status": "waiting"},
+        {
+            "timestamp": "2019-01-01 00:00:00.1",
+            "equilibration status": "waiting",
+            "setpoint temperature": 40,
+        },
+        {
+            "timestamp": "2019-01-01 00:00:01.1",
+            "equilibration status": "equilibrated",
+            "setpoint temperature": 40,
+        },
+        {
+            "timestamp": "2019-01-01 00:00:03.1",
+            "equilibration status": "equilibrated",
+            "setpoint temperature": 40,
+        },
+        {
+            "timestamp": "2019-01-01 00:00:04.1",
+            "equilibration status": "waiting",
+            "setpoint temperature": 40,
+        },
     ]
 )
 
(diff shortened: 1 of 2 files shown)


jason-curtis (approved):

rush340 (approved):

osmo_jupyter/dataset_test.py
```diff
@@ -408,6 +408,7 @@ def test_filters_all_data_to_equilibrated_states(
                         "ROI 0 g_msorm": 0.6,
                         "ROI 1 g_msorm": 0.3,
                         "image": "image-1.jpeg",
+                        "PicoLog Temperature Ave. (C)": 39.75,
```
jaimemarijke: is this smoke-testing that the actual interpolation works correctly? or does the timestamp happen to line up in this test?

jaimemarijke (approved):
Could be good to add a test of the interpolation (including interesting edge cases), if the existing test doesn't cover that.

osmo_jupyter/dataset_test.py
EvanSimpson: I'm not sure what you mean by "happen to line up" but this is the actual interpolated value.
It's also sort of covering an interesting (edge?) case, which is that the image was taken at a timestamp present in the PicoLog data, but not the calibration data.

The flow of the data is kind of like this - the initial raw sensor data, shown sparsely here:

| Second | PicoLog Temp | Calibration Status |
| --- | --- | --- |
| 00 |  39 | waiting |
| 01 | -- | equilibrated |
| 02 |  40 | -- |
| 03 | --  | equilibrated |
| 04 | 40 | waiting |

Then the PicoLog data is upsampled to each second and linearly interpolated. Then the two are joined together on matching timestamps, essentially any timestamp present in the calibration data:

| Second | Calibration Status | Interpolated Temperature |
| --- | --- | --- |
| 00 | waiting | 39 |
| 01 | equilibrated | 39.5 |
| 03 | equilibrated | 40 |
| 04 | waiting | 40 |

Then when that dataframe is joined on the ROI data, it's now being upsampled and linearly interpolated again:

| Second | Calibration Status | Re-interpolated Temperature |
| --- | --- | --- |
| 00 | waiting | 39 |
| 01 | equilibrated | 39.5 |
| 02 | equilibrated | 39.75 |
| 03 | equilibrated | 40 |
| 04 | waiting | 40 |

It's probably also important to point out that a 1 degree change over 1 second is a large exaggeration so a discrepancy of .25 degrees between the original and re-interpolated values is probably unrealistic, especially considering all of this data is being filtered down to equilibrated regions.


I feel like it's worth adding _some_ of this as documentation in the code, but I'm not sure how much or where...

jaimemarijke: Wow, yes this is an excellent description, thank you. 

I agree that it would be nice to have something like this recorded in the code, and I'm also unsure where it would belong. Options I can think of:

0. Nowhere. 
1. A module-level docstring. Would give a nice overview but probably go instantly out of date... 
2. Interspersed in the function. Probably harder too grok but less likely to get out of date.

jaimemarijke: Also, this does make me wonder - would it work to join the PicoLog data to the ROI data first, so that if there does happen to be a temperature reading at the exact same time as the image, that data isn't lost to interpolation? Admittedly, it seems like the error would be small.

jason-curtis: Hmm, I guess we could technically avoid that and get potentially more correct results by upsampling _both_ the picolog and calibration data, joining those, and then joining those with the ROI data. Given that we have:

* picolog data every M(?) seconds
* calibration data every N(9) seconds
* ROI data every J(9) seconds

What's the worst case? With some handwavey math, it seems like maybe the picolog data could be off by something like M/2 + N/2 seconds?

EvanSimpson: I am wary of interpolating the calibration data before pulling out the equilibration boundaries. That puts us back at that convoluted method of trimming the interpolated edges I had before. 
I think I might go with Jaime's suggestion of joining the PicoLog and ROI data independently first, since the ROI doesn't need to be interpolated.

PR #35, opened 2019-10-18
Add function to open and clean up any data file

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1144794081013111)

Adds a helper function that can identify and open any of our typical data files from calibration or scum tank data collection. Also cleans up column names and timestamps to be standardized across files, mostly using the calibration log files as the baseline.

The way it's structured is such that the actual opening of the files is done in separate functions, and there's one wrapper function which will attempt to infer the filetype (or you can skip that and just tell it) and then apply necessary transformations. The type inference feels kind of fragile, and like an over-the-top convenience, so I'd also be happy cutting it out.

## Manual Testing
Ran this code with real files and checked the output:
```python
files = [
    "YSI_data.csv",
    "2019-10-09_pico.csv",
    "2019-10-11--19-41-03_calibration.csv",
    "KorDSS Measurement File Export - 101119 214237.csv",
]
for file in files:
    display(read_and_format_data_file(file).head())
```

### DIFF
--- a/osmo_jupyter/dataset.py
+++ b/osmo_jupyter/dataset.py
@@ -1,7 +1,175 @@
 import os
-from typing import List
+from typing import List, Dict
+from enum import Enum
 
 import pandas as pd
+
+
+# Standard column names to align various data formats on
+TIMESTAMP = "timestamp"
+TEMPERATURE_C = "temperature (C)"
+BAROMETRIC_PRESSURE_MMHG = "barometric pressure (mmHg)"
+DO_PCT = "DO (%)"
+DO_MGL = "DO (mg/L)"
+
+
+class DataFileType(Enum):
+    UNKNOWN = 0
+    YSI_CSV = 1
+    YSI_KORDSS = 2
+    PICOLOG = 3
+    CALIBRATION_LOG = 4
+
+
+def read_ysi_kordss_file(filepath: str) -> pd.DataFrame:
+    """ Open a YSI KorDSS formatted csv file, with standardized datetime column parsing.
+
+        Args:
+            filepath: Filepath to a YSI KorDSS csv file.
+        Returns:
+            Pandas DataFrame of the raw data, with DATE and TIME columns
+            parsed together into a DATE_TIME column with a datetime dtype.
+    """
+    return pd.read_csv(
+        filepath, skiprows=5, encoding="latin-1", parse_dates=[["DATE", "TIME"]]
+    )
+
+
+def read_ysi_csv_file(filepath: str) -> pd.DataFrame:
+    """ Open a YSI "classic" csv file, with standardized datetime column parsing.
+
+        Args:
+            filepath: Filepath to a YSI csv file.
+        Returns:
+            Pandas DataFrame of the raw data, with Timestamp column parsed as a datetime dtype.
+    """
+    return pd.read_csv(filepath, parse_dates=["Timestamp"])
+
+
+def read_picolog_file(filepath: str) -> pd.DataFrame:
+    """ Open a PicoLog csv file, with standardized datetime column parsing.
+
+        Args:
+            filepath: Filepath to a PicoLog csv file.
+        Returns:
+            Pandas DataFrame of the raw data, with the unlabeled timestamp column parsed
+            as a datetime dtype with the timezone stripped.
+    """
+    return pd.read_csv(
+        filepath,
+        parse_dates=[0],
+        date_parser=lambda col: pd.to_datetime(col, utc=False).tz_localize(None),
+    )
+
+
+def read_calibration_log_file(filepath: str) -> pd.DataFrame:
+    """ Open a calibration log csv file, with standardized datetime column parsing.
+
+        Args:
+            filepath: Filepath to a PicoLog csv file.
+        Returns:
+            Pandas DataFrame of the raw data, with timestamp column parsed
+            as a datetime dtype with fractional seconds truncated.
+    """
+    return pd.read_csv(
+        filepath,
+        parse_dates=["timestamp"],
+        date_parser=lambda col: pd.to_datetime(col, utc=False).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        ),  # Truncate fractional seconds
+    )
+
+
+# Configure how each type of file we handle should be opened and transformed
+FILE_FORMAT_CONFIG = {
+    DataFileType.YSI_CSV: {
+        "rename": {
+            "Timestamp": TIMESTAMP,
+            "Barometer (mmHg)": BAROMETRIC_PRESSURE_MMHG,
+            "Dissolved Oxygen (%)": DO_PCT,
+            "Temperature (C)": TEMPERATURE_C,
+            "Unit ID": "unit ID",
+        },
+        "drop": ["Comment", "Site", "Folder"],
+        "prefix": "YSI ",
+        "parser": read_ysi_csv_file,
+    },
+    DataFileType.YSI_KORDSS: {
+        "rename": {
+            "DATE_TIME": TIMESTAMP,
+            "Barometer (mmHg)": BAROMETRIC_PRESSURE_MMHG,
+            "ODO (% Sat)": DO_PCT,
+            "ODO (mg/L)": DO_MGL,
+            "Temp (°C)": TEMPERATURE_C,
+        },
+        "drop": ["SITE", "DATA ID", "ODO (% Local)"],
+        "prefix": "YSI ",
+        "parser": read_ysi_kordss_file,
+    },
+    DataFileType.PICOLOG: {
+        "rename": {
+            "Unnamed: 0": TIMESTAMP,
+            "Temperature Ave. (C)": TEMPERATURE_C,
+            "Pressure Ave. (mmHg)": BAROMETRIC_PRESSURE_MMHG,
+        },
+        "drop": ["Pressure (Voltage) Ave. (nV)"],
+        "prefix": "PicoLog ",
+        "parser": read_picolog_file,
+    },
+    DataFileType.CALIBRATION_LOG: {
+        "rename": {},
+        "drop": [],
+        "prefix": "",
+        "parser": read_calibration_log_file,
+    },
+}
+
+
+def _infer_filetype(filename: str) -> str:
+    with open(filename, "r") as file:
+        try:
+            header = file.readline()
+            if "Dissolved Oxygen (%)" in header:
+                return DataFileType.YSI_CSV
+            elif "Pressure (Voltage) Ave." in header:
+                return DataFileType.PICOLOG
+            elif "O2 source gas gas ID" in header:
+                return DataFileType.CALIBRATION_LOG
+        except UnicodeDecodeError:
+            return DataFileType.YSI_KORDSS
+    return DataFileType.UNKNOWN
+
+
+def _apply_parser_configuration(
+    dataset: pd.DataFrame, parse_config: Dict
+) -> pd.DataFrame:
+    return (
+        dataset.drop(parse_config["drop"], axis=1)
+        .rename(columns=parse_config["rename"])
+        .set_index("timestamp")
+        .add_prefix(parse_config["prefix"])
+    )
+
+
+def read_and_format_data_file(
+    filepath: str, filetype: DataFileType = DataFileType.UNKNOWN
+) -> pd.DataFrame:
+    """
+        Open and format any of the following supported data file types with standardized column names:
+        Original YSI csv, KorDSS YSI csv, PicoLog csv, calibration log csv
+
+        Args:
+            filepath: Filepath to the file to be opened and formatted.
+            filetype: Optional. Provide the DataFileType if it is known. Otherwise the filetype will be inferred.
+    """
+    if filetype is DataFileType.UNKNOWN:
+        filetype = _infer_filetype(filepath)
+    parse_config = FILE_FORMAT_CONFIG[filetype]
+    parse_function = parse_config["parser"]
+
+    dataset = parse_function(filepath)
+
+    return _apply_parser_configuration(dataset, parse_config)
 
 
 def open_picolog_file(picolog_filepath: str) -> pd.DataFrame:
@@ -15,16 +183,7 @@ def open_picolog_file(picolog_filepath: str) -> pd.DataFrame:
             upsampled and linearly interpolated to each intermediate second.
     """
     return (
-        pd.read_csv(
-            picolog_filepath,
-            index_col=0,  # first column, is unlabeled
-            parse_dates=[0],  # first column
-            date_parser=lambda col: pd.to_datetime(col, utc=False).tz_localize(
-                None
-            ),  # Remove timezone information
-        )
-        .rename_axis("timestamp")
-        .add_prefix("PicoLog ")
+        read_and_format_data_file(picolog_filepath, DataFileType.PICOLOG)
         .resample("s")
         .interpolate(method="slinear")
     )
@@ -39,13 +198,8 @@ def open_calibration_log_file(calibration_log_filepath: str) -> pd.DataFrame:
         Returns:
             DataFrame of calibration data file contents, with timestamp indexes.
     """
-    return pd.read_csv(
-        calibration_log_filepath,
-        index_col="timestamp",
-        parse_dates=["timestamp"],
-        date_parser=lambda col: pd.to_datetime(col, utc=False).strftime(
-            "%Y-%m-%d %H:%M:%S"
-        ),  # Truncate fractional seconds
+    return read_and_format_data_file(
+        calibration_log_filepath, DataFileType.CALIBRATION_LOG
     )
 
 
--- a/osmo_jupyter/dataset.py
+++ b/osmo_jupyter/dataset.py
@@ -33,13 +33,14 @@ def _apply_parser_configuration(
 
 
 def parse_ysi_kordss_file(filepath: str) -> pd.DataFrame:
-    """ Open a YSI KorDSS formatted csv file, with standardized datetime column parsing.
+    """ Open and format a YSI KorDSS formatted csv file, with standardized datetime parsing
+        and cleaned up columns.
 
         Args:
             filepath: Filepath to a YSI KorDSS csv file.
         Returns:
-            Pandas DataFrame of the raw data, with DATE and TIME columns
-            parsed together into a DATE_TIME column with a datetime dtype.
+            Pandas DataFrame of the data, with DATE and TIME columns parsed together,
+            and standardized column names.
     """
     parse_config = {
         "rename": {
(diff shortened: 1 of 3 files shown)


osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+            elif "Pressure (Voltage) Ave." in header:
+                return DataFileType.PICOLOG
+            elif "O2 source gas gas ID" in header:
+                return DataFileType.CALIBRATION_LOG
+        except UnicodeDecodeError:
+            return DataFileType.YSI_KORDSS
+    return DataFileType.UNKNOWN
+
+
+def _apply_parser_configuration(
```
EvanSimpson: This is for easier testing. Since the test file already had some helper code to do everything in-memory rather than with temp files, I stuck with that approach, which meant the file type inference was incompatible, so I focused on testing the individual `read_*_file` functions and this one separately from  `read_and_format_data_file`.

Of course the KorDSS file ended up requiring a temp file to test anyway.

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+    with open(filename, "r") as file:
+        try:
+            header = file.readline()
+            if "Dissolved Oxygen (%)" in header:
+                return DataFileType.YSI_CSV
+            elif "Pressure (Voltage) Ave." in header:
+                return DataFileType.PICOLOG
+            elif "O2 source gas gas ID" in header:
+                return DataFileType.CALIBRATION_LOG
+        except UnicodeDecodeError:
```
EvanSimpson: 💃 `latin-1` 💃

osmo_jupyter/dataset.py
```diff
@@ -15,16 +183,7 @@ def open_picolog_file(picolog_filepath: str) -> pd.DataFrame:
...
-            picolog_filepath,
-            index_col=0,  # first column, is unlabeled
-            parse_dates=[0],  # first column
-            date_parser=lambda col: pd.to_datetime(col, utc=False).tz_localize(
-                None
-            ),  # Remove timezone information
-        )
-        .rename_axis("timestamp")
-        .add_prefix("PicoLog ")
+        read_and_format_data_file(picolog_filepath, DataFileType.PICOLOG)
```
EvanSimpson: I left this and the following functions so as not to cause chaos in some notebooks that might already use them, but this was a nice refactor.

osmo_jupyter/dataset_test.py
```diff
@@ -122,25 +282,25 @@ def test_interpolates_data_correctly(
                     "timestamp": "2019-01-01 00:00:00",
                     "equilibration status": "waiting",
                     "setpoint temperature": 40,
-                    "PicoLog Temperature Ave. (C)": 39,
+                    "PicoLog temperature (C)": 39,
```
EvanSimpson, line 253: These are actually being cleaned up now since I changed `open_picolog_file` to use the new formatting function.

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+        parse_dates=["timestamp"],
+        date_parser=lambda col: pd.to_datetime(col, utc=False).strftime(
+            "%Y-%m-%d %H:%M:%S"
+        ),  # Truncate fractional seconds
+    )
+
+
+# Configure how each type of file we handle should be opened and transformed
+FILE_FORMAT_CONFIG = {
+    DataFileType.YSI_CSV: {
```
jaimemarijke: huh, I was expecting this to happen inside the `read_ysi_csv_file` function. Is there a reason you didn't do it there?

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+        Returns:
+            Pandas DataFrame of the raw data, with DATE and TIME columns
+            parsed together into a DATE_TIME column with a datetime dtype.
+    """
+    return pd.read_csv(
+        filepath, skiprows=5, encoding="latin-1", parse_dates=[["DATE", "TIME"]]
+    )
+
+
+def read_ysi_csv_file(filepath: str) -> pd.DataFrame:
```
jaimemarijke: nit: maybe this should be called `read_ysi_classic_file` or similar, since being a "csv" doesn't actually distinguish it from the other type.

osmo_jupyter/dataset.py
jaimemarijke: ah, maybe it was to DRY up applying the renaming/dropping/prefixing actions?

I think this is verging to me towards being so DRY as to be parched, and I would slightly prefer the `read_*_file` functions to just handle everything that is unique to a file type, so that they each return an appropriately reformatted df.

osmo_jupyter/dataset.py
jaimemarijke: since our pattern has been to have notebooks pin to a precise git commit, I would be okay with just removing the old stuff.

osmo_jupyter/dataset.py
jaimemarijke: probably should have a comment explaining that

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+        Open and format any of the following supported data file types with standardized column names:
+        Original YSI csv, KorDSS YSI csv, PicoLog csv, calibration log csv
+
+        Args:
+            filepath: Filepath to the file to be opened and formatted.
+            filetype: Optional. Provide the DataFileType if it is known. Otherwise the filetype will be inferred.
+    """
+    if filetype is DataFileType.UNKNOWN:
+        filetype = _infer_filetype(filepath)
+    parse_config = FILE_FORMAT_CONFIG[filetype]
```
jaimemarijke: looks like this will raise with a key error if filetype is UNKNOWN and `_infer_filetype` can't figure it out. Maybe something should instead explicitly handle that case and raise with a more helpful error message.

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+) -> pd.DataFrame:
+    return (
+        dataset.drop(parse_config["drop"], axis=1)
+        .rename(columns=parse_config["rename"])
+        .set_index("timestamp")
+        .add_prefix(parse_config["prefix"])
+    )
+
+
+def read_and_format_data_file(
```
jaimemarijke: nit: I might just use the word "parse" instead of "read_and_format".. I think that's more or less what "parse" means to me

osmo_jupyter/dataset_test.py
```diff
@@ -73,22 +116,139 @@
...
 def mock_calibration_file_obj():
-    test_calibration_env_file = io.StringIO()
-    TEST_CALIBRATION_DATA.to_csv(test_calibration_env_file, index=False)
-    test_calibration_env_file.seek(0)
+    return create_mock_file_obj(TEST_CALIBRATION_DATA)
+
+
+class TestReadAndFormatDataFile:
+    def test_formats_ysi_csv_correctly(self):
+        mock_YSI_csv_file_obj = create_mock_file_obj(TEST_YSI_CSV_DATA)
```
jaimemarijke: huh, I'm kind of confused what's happening here. Doesn't the `read_*_file` function expect a filepath, not a file object?

osmo_jupyter/dataset.py
jaimemarijke: maybe it would be better to just switch everything to use temp files?

osmo_jupyter/dataset.py
jaimemarijke: (also, if you end up moving more formatting logic into the `read_*_file` functions, maybe they should also be renamed to `parse_*_file`)

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+            filetype: Optional. Provide the DataFileType if it is known. Otherwise the filetype will be inferred.
+    """
+    if filetype is DataFileType.UNKNOWN:
+        filetype = _infer_filetype(filepath)
+    parse_config = FILE_FORMAT_CONFIG[filetype]
+    parse_function = parse_config["parser"]
+
+    dataset = parse_function(filepath)
+
+    return _apply_parser_configuration(dataset, parse_config)
```
jaimemarijke: minor: seeing how this plays out down here, I'm feeling more strongly that it has crossed over into being so DRY as to be parched, and I'm thinking it might be nicer if the `read_*_file` functions did more of the heavy lifting, and this function really just picked which of those functions to use based on the DataFileType

osmo_jupyter/dataset_test.py
jaimemarijke: ah, it looks `pd.read_csv` just happens to handle being passed either a filepath or a buffer. 
I don't love this test then, since it's not quite testing the real-life scenario. I'd prefer if these tests were just refactored to use temp files. Maybe when we wrote this code we hadn't figured out how easy it is to use the pytest `tmp_path` fixture?

osmo_jupyter/dataset.py
```diff
@@ -1,9 +1,177 @@
...
+    DataFileType.CALIBRATION_LOG: {
+        "rename": {},
+        "drop": [],
+        "prefix": "",
+        "parser": read_calibration_log_file,
+    },
+}
+
+
+def _infer_filetype(filename: str) -> str:
```
jaimemarijke: you mentioned that maybe this is too brittle / an over-the-top convenience and I think I agree. Seems like we should always know what type the file we're trying to open is, and if we don't we probably have other problems.

jaimemarijke (commented):
I'm somewhere between "approve" and "request changes". I feel moderately strongly that some refactoring would be helpful, especially of the tests.

osmo_jupyter/dataset.py
EvanSimpson: Done, feels a lot cleaner now.

osmo_jupyter/dataset.py
EvanSimpson: 💣 💥

osmo_jupyter/dataset.py
EvanSimpson: I refactored all these read_ functions to be parse_ and moved all the individual configurations to be internal.

osmo_jupyter/dataset_test.py
EvanSimpson: Done.

osmo_jupyter/dataset.py
EvanSimpson: I ended up deleting this code and the top-level function as well, since if you're going to pass if the dtype anyway, you might as well just use the right function.

osmo_jupyter/dataset.py
EvanSimpson: Removed this code so no longer relevant.

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
 import os
```
jason-curtis, line 1: feels like it would be natural for this to have its own module, `(load_)source_data`?

osmo_jupyter/dataset_test.py
```diff
@@ -73,41 +114,125 @@
...
+                    "YSI unit ID": "unit ID",
+                }
+            ]
+        ).set_index("timestamp")
+
+        pd.testing.assert_frame_equal(formatted_ysi_data, expected_ysi_data)
+
+    def test_formats_ysi_kordss_correctly(self, tmpdir):
+        mock_YSI_KorDSS_file_obj = tmpdir.join("test_kordss.csv")
+        mock_YSI_KorDSS_file_obj.write("\n\n\n\n\n")
```
jason-curtis:
```suggestion
        # Real KorDSS file has other contents in this header, but it adds up to 5 lines
        mock_YSI_KorDSS_file_obj.write("\n\n\n\n\n")
```

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
...
+
+
+class DataFileType(Enum):
+    YSI_CLASSIC = 0
+    YSI_KORDSS = 1
+    PICOLOG = 2
+    CALIBRATION_LOG = 3
+
+
+def _apply_parser_configuration(
```
jason-curtis, line 15: nice abstraction :+1:

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
...
+            Pandas DataFrame of the raw data, with timestamp column parsed
+            as a datetime dtype with fractional seconds truncated.
+    """
+    parse_config = {"rename": {}, "drop": [], "prefix": ""}
+    raw_data = pd.read_csv(
+        filepath,
         parse_dates=["timestamp"],
         date_parser=lambda col: pd.to_datetime(col, utc=False).strftime(
             "%Y-%m-%d %H:%M:%S"
         ),  # Truncate fractional seconds
```
jason-curtis, line 122: :thinking: very interesting that truncation is the right thing to do here, and not rounding (since everything else is - presumably? - truncating when reporting full seconds). Guess every rule has an exception :man_shrugging:

jason-curtis (approved):

osmo_jupyter/dataset.py
jason-curtis: eh, never mind, it's really in line with what's in there already, though that file might be just about due for being spilt up.

EvanSimpson: Yeah, I can revisit this with the addition of the data collection log. Might be better, though I am a little worried about losing some shared code in the tests, and importing helper code from another test file seems awkward...

osmo_jupyter/dataset.py
jaimemarijke: hah, I'm torn on whether I'd prefer just performing these pandas df operations directly in the `parse_*_file` functions, since this doesn't actually do all that much more than just calling the pandas stuff directly. But I guess it does form an interface of sorts, e.g. ensuring everyone sets the index to the timestamp. 

It did strike me as slightly odd that the `parse_calibration_log_file` function had to pass an empty config just to get the set_index functionality - made me wonder if maybe a better interface would be named kwargs with empty defaults instead of a `parse_config` dict. 

But, this is all getting pretty nitpicky and personal-preference-y. I think this refactoring is fine as-is.

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
...
+    YSI_KORDSS = 1
+    PICOLOG = 2
+    CALIBRATION_LOG = 3
+
+
+def _apply_parser_configuration(
+    dataset: pd.DataFrame, parse_config: Dict
+) -> pd.DataFrame:
+    return (
+        dataset.drop(parse_config["drop"], axis=1)
```
jaimemarijke: nit: looks like `drop` accepts a `columns` kwargs, which I personally find easier to parse than `axis=1` and would also match the line below.

osmo_jupyter/dataset_test.py
```diff
@@ -73,41 +114,126 @@
...
-def mock_picolog_file_obj():
-    test_picolog_file = io.StringIO()
-    TEST_PICOLOG_DATA.to_csv(test_picolog_file, index=False)
-    test_picolog_file.seek(0)
+def create_mock_file_obj(tmpdir, data, file_name):
+    mock_file_obj = tmpdir.join(file_name)
+    data.to_csv(mock_file_obj, index=False)
 
-    return test_picolog_file
+    return mock_file_obj
```
jaimemarijke: is this the file object or the filename? I would have thought it would be the filename

jaimemarijke: looks like it is a Path object: http://doc.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture
so probably this should just be renamed to `mock_file_path` or `tmp_file_path` or something (here and elsewhere)

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
...
-        Open and parse a PicoLog data file such that the timestamps can be joined with calibration environment data.
+# Standard column names to align various data formats on
+TIMESTAMP = "timestamp"
+TEMPERATURE_C = "temperature (C)"
+BAROMETRIC_PRESSURE_MMHG = "barometric pressure (mmHg)"
+DO_PCT = "DO (%)"
+DO_MGL = "DO (mg/L)"
+
+
+class DataFileType(Enum):
```
jaimemarijke: minor: looks like this is unused now

osmo_jupyter/dataset_test.py
```diff
@@ -73,41 +114,126 @@
...
+        tmpdir, TEST_CALIBRATION_DATA, "test_calibration_data.csv"
+    )
 
-class TestOpenAndCombineSensorData:
-    def test_parses_picolog_timestamps_correctly(self, mock_picolog_file_obj):
-        picolog_data = module.open_picolog_file(mock_picolog_file_obj)
 
-        assert len(picolog_data)
-        assert picolog_data.index[0] == pd.to_datetime("2019-01-01 00:00:00")
+class TestReadAndFormatDataFile:
```
jaimemarijke: I think the `read_and_format_data_file` function doesn't exist at all any more. Could just remove the class, since looks like each of the file-specific functions only has one test. Or group each of the test functions under their own class.

osmo_jupyter/dataset_test.py
```diff
@@ -73,41 +114,126 @@
 )
 
 
-@pytest.fixture
-def mock_picolog_file_obj():
-    test_picolog_file = io.StringIO()
-    TEST_PICOLOG_DATA.to_csv(test_picolog_file, index=False)
-    test_picolog_file.seek(0)
+def create_mock_file_obj(tmpdir, data, file_name):
```
jaimemarijke: do you have to pass the tmpdir fixture to this explicitly? doesn't pytest just magically stuff it in if you don't? or is that only in test functions?

jaimemarijke (approved):

osmo_jupyter/dataset.py
```diff
@@ -1,53 +1,137 @@
...
 import pandas as pd
 
 
-def open_picolog_file(picolog_filepath: str) -> pd.DataFrame:
-    """
-        Open and parse a PicoLog data file such that the timestamps can be joined with calibration environment data.
+# Standard column names to align various data formats on
+TIMESTAMP = "timestamp"
+TEMPERATURE_C = "temperature (C)"
+BAROMETRIC_PRESSURE_MMHG = "barometric pressure (mmHg)"
```
rush340, line 10: it looks like we rename this to "barometer (mmHg)" in our datasets, should we go directly to that name here?

rush340 (approved):

osmo_jupyter/dataset_test.py
EvanSimpson: I think they will still need their own class. I seem to remember not being able to mix class and non-class based tests in the same file.

osmo_jupyter/dataset_test.py
EvanSimpson: That's only in tests and fixtures.

osmo_jupyter/dataset.py
EvanSimpson: Hmm. Good point. This change was intended to match thing up to the calibration log, but I guess that was being changed in the datasets as well. I can change it here, but we should also think about changing it at the source in the calibration YSI driver as well.

EvanSimpson: Thinking about this some more, I think I prefer `"barometric pressure"` to `"barometer"`, since we have `"temperature"` and `"timestamp"` rather than `"thermometer"` and `"clock"`.

osmo_jupyter/dataset_test.py
EvanSimpson: Nevermind, don't know where I got that in my head.

osmo_jupyter/dataset.py
rush340: yeah, I also prefer "barometric pressure". Maybe it's better to change downstream consumers and get rid of that renaming at some point.

EvanSimpson: I don't know that the barometric pressure in the dataset files is actually used anywhere currently, so we could probably check that one off.

PR #36, opened 2019-10-22
Add data collection log parser and refactor dataset tools

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1144794081013110)

* Splits up the dataset module into `parse` and `combine` submodules.
* Adds a helper function to parse the anglerfish data collection log
    * Returns a data frame with a subset of columns (`experiment_names`, `drive_directory`, `pond`, `cosmobot_id`, `start_date`, `end_date`)
    * `pond` kind of means "environment" unless we consider the calibration bath a pond. I'm not totally sold on that name, but don't know that "environment" is any more accurate.
    * Uses `pd.read_excel` so the log must be exported from google as a `.xlsx`. This limited the options for testing to committing an actual test file.

## Manual Testing:
Exported the current data collection log as a `.xlsx` file, looked at the dataframe output.


jason-curtis (approved):
lgtm.
somewhat on the fence on the re-exporting in `__init__.py`. Do you think we want to export everything at the top level even as we add more things? I kinda want to just rip the band-aid and include nothing or almost-nothing in `__init__` but could go either way.

osmo_jupyter/dataset/combine_test.py
```diff
@@ -1,125 +1,16 @@
...
-            "Dissolved Oxygen (%)": 19,
-            "Temperature (C)": 24.7,
-            "Unit ID": "unit ID",
-            "Comment": None,
-            "Site": None,
-            "Folder": None,
-        }
-    ]
+import osmo_jupyter.dataset.combine as module
+from osmo_jupyter.dataset.test_fixtures import (  # noqa: F401
```
jaimemarijke: nit: missing a description of what flake rule is being ignored

osmo_jupyter/dataset/combine_test.py
```diff
@@ -1,125 +1,16 @@
...
+    create_mock_file_path,
+    TEST_YSI_CSV_DATA,
```
jaimemarijke: nit: maybe this should be renamed to `TEST_YSI_CLASSIC_DATA`

osmo_jupyter/dataset/parse.py
```diff
@@ -0,0 +1,176 @@
...
+        }
+    )
+
+
+def parse_data_collection_log(filepath: str) -> pd.DataFrame:
+    """ Open and summarize a data collection log .xlsx file containing sheets named
+        "Calibration Environment" and "Scum tank".
+
+        Args:
+            filepath: Filepath to a data collection log .xlsx file
```
jaimemarijke, line 147: woah does Google Sheets use .xlsx? or is the idea that we'd export the google sheet as .xlsx? Seems better to stick with csv if we have that option - maybe we'd have to export the two sheets separately though?

osmo_jupyter/dataset/test_fixtures.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def create_mock_file_path(tmpdir, data, file_name):
+    mock_file_path = tmpdir.join(file_name)
+    data.to_csv(mock_file_path, index=False)
+
+    return mock_file_path
+
+
+TEST_YSI_CSV_DATA = pd.DataFrame(
```
jaimemarijke: at the point where we have to pull these out into a separate `test_fixtures` file, I'm leaning towards just creating an `test_fixtures` folder with actual files in it. Actually, that might be an appropriate thing to do here in any case since the data files themselves constitute an external interface of sorts - it would have the benefit of having concrete examples of what the files actually look like (e.g. the header in the KorDSS file)

jaimemarijke (approved):
The refactor seems like a good move. 
As I commented below, I'm thinking maybe the test fixtures should just be actual files in a `test_fixtures` folder.

rush340 (approved):

osmo_jupyter/dataset/parse.py
EvanSimpson: The idea is that we would export one xlsx file to get both sheets in one file. Csv is an option, but yes, requires separate files.

osmo_jupyter/dataset/test_fixtures.py
EvanSimpson: Went through and made this change.

osmo_jupyter/dataset/parse_test.py
```diff
@@ -71,60 +47,56 @@ def test_parses_ysi_kordss_correctly(tmpdir):
...
+            pd.to_datetime("2019-01-01 00:00:01"),
+            pd.to_datetime("2019-01-01 00:00:03"),
+            pd.to_datetime("2019-01-01 00:00:04"),
+        ],
+        name="timestamp",
+    )
 
-    pd.testing.assert_frame_equal(
-        formatted_calibration_log_data, expected_calibration_log_data
+    pd.testing.assert_index_equal(
```
jaimemarijke, line 98: should this not still test that other columns were pulled in correctly?

jaimemarijke (approved):
nice, I do prefer seeing the what the actual data files look like.

osmo_jupyter/dataset/parse_test.py
EvanSimpson: I thought about that, and figured that would really just be testing that pandas' `read_csv` was working, since parsing the timestamp and setting it as the index are the only transforms happening on this data. Open to stronger opinions about whether we should be checking all or some of the other 20-something columns.

jason-curtis (approved):
:+1: I also like having actual file specimens!

rush340 (approved):

PR #38, opened 2019-11-05
Get experiment data files from Google Drive format

jason-curtis:
[asana](https://app.asana.com/0/819671808102776/1147515614239923/f)

We talked about leaving this in a notebook for this ticket but it seemed obvious enough where it should go in a repo.

## Manual testing

Used on all the Drive directories in the log and eyeballed the results. see my fork of the Parse and combine calibration and sensor data snippet: http://marvin:8889/notebooks/Tools/Jupyter%20Snippets/Parse%20and%20combine%20calibration%20and%20sensor%20data-Temp%20version%20updated%20with%20Jason%20code.ipynb

this is what it took to tack the files on to a parsed data collection log:
```
from osmo_jupyter.dataset.files import get_experiment_data_files_by_type

DRIVE_EXPERIMENTS_ROOT = "/data/Google Drive/Technical/Experiments"
drive_directory_paths = parsed_log["drive_directory"].apply(
    lambda directory_name: os.path.join(DRIVE_EXPERIMENTS_ROOT, directory_name)
)
parsed_log_with_file_paths = pd.concat(
    [
        parsed_log, # parsed using `parse_data_collection_log`
        drive_directory_paths.apply(get_experiment_data_files_by_type).add_suffix(
            "_files"
        ),
    ],
    axis="columns",
)
```
image

*suggestion: I should probably go ahead and librarify this as well*, probably as `osmo_jupyter.dataset.parse.parse_data_collection_log_with_drive_files(log_filepath, drive_experiments_root)` but I'm not sure how much to pull on this scope creep thread before I'm stepping on Evan's ticket's toes! The logical next step after that would be to modify/provide a version of `open_and_combine_and_filter_source_data()` that's integrated with this also.

### DIFF
--- a/osmo_jupyter/dataset/files.py
+++ b/osmo_jupyter/dataset/files.py
@@ -24,11 +24,11 @@ def _get_experiment_data_file_paths_for_type(project_directory, file_type):
         os.path.join(subdirectory_path, filename)
         for filename in os.listdir(subdirectory_path)
     ]
-    files_in_subdirectory = [
+    files_in_subdirectory = sorted(
         filepath
         for filepath in files_and_folders_in_subdirectory
         if os.path.isfile(filepath)
-    ]
+    )
 
     return files_in_subdirectory
 
--- a/osmo_jupyter/dataset/files.py
+++ b/osmo_jupyter/dataset/files.py
@@ -1,0 +1,54 @@
+import os
+import pandas as pd
+
+FILE_TYPES = [
+    "setpoints",
+    "calibration_log",
+    "pico",
+    "process_experiment",
+    "ysi_prodo",
+    "ysi_prosolo",
+    "summary_movies",
+]
+
+DATA_DIRECTORY_NAME = "data"
+
+
+def _get_experiment_data_file_paths_for_type(project_directory, file_type):
+    subdirectory_path = os.path.join(project_directory, DATA_DIRECTORY_NAME, file_type)
+
+    if not os.path.exists(subdirectory_path):
+        return []
+
+    files_and_folders_in_subdirectory = [
+        os.path.join(subdirectory_path, filename)
+        for filename in os.listdir(subdirectory_path)
+    ]
+    files_in_subdirectory = [
+        filepath
+        for filepath in files_and_folders_in_subdirectory
+        if os.path.isfile(filepath)
+    ]
+
+    return files_in_subdirectory
+
+
+def get_experiment_data_files_by_type(project_directory):
+    """ Spider the provided directory for files related to an experiment.
+    Expects the project directory to match our standard Google Drive experiment directory format.
+
+    Args:
+        project_directory: Google Drive experiment directory containing all of
+            the data files for this data collection attempt
+    Returns:
+        pandas Series, indexed by file type and containing lists of filenames.
+    """
+
+    return pd.Series(
+        {
+            file_type: _get_experiment_data_file_paths_for_type(
+                project_directory, file_type
+            )
+            for file_type in FILE_TYPES
+        }
+    )
--- a/osmo_jupyter/dataset/files.py
+++ b/osmo_jupyter/dataset/files.py
@@ -6,7 +6,7 @@ FILE_TYPES = [
     "calibration_log",
     "pico",
     "process_experiment",
-    "ysi_prodo",
+    "ysi_proodo",
     "ysi_prosolo",
     "summary_movies",
 ]
--- a/osmo_jupyter/dataset/files_test.py
+++ b/osmo_jupyter/dataset/files_test.py
@@ -34,22 +34,22 @@ class TestGetExperimentDataFilePathsForType:
     def test_gets_file_from_subdirectory_and_ignores_sub_subdirectories(self, tmp_path):
         _init_data_dir(
             tmp_path,
-            directories_to_include=["ysi_prodo/subdirectory"],
-            files_to_include=["ysi_prodo/somethin"],
+            directories_to_include=["ysi_proodo/subdirectory"],
+            files_to_include=["ysi_proodo/somethin"],
         )
-        actual = module._get_experiment_data_file_paths_for_type(tmp_path, "ysi_prodo")
-        expected = [os.path.join(tmp_path, "data/ysi_prodo/somethin")]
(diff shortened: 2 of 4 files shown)


EvanSimpson (approved):
I'd be fine with pushing the scope creep back and librarifying the usage of this code in a future ticket, once we're ready to build the end-to-end pipeline.

osmo_jupyter/dataset/files.py
```diff
@@ -0,0 +1,54 @@
+import os
+import pandas as pd
+
+FILE_TYPES = [
```
jaimemarijke: are each of these their own folder? that's not what I expected from the name `FILE_TYPES`, but that's what it seems like they must be based on line 18

osmo_jupyter/dataset/files_test.py
```diff
@@ -0,0 +1,87 @@
...
+    data_dir = os.path.join(parent_dir, "data")
+    os.mkdir(data_dir)
+
+    for relative_directory_path in directories_to_include:
+        os.makedirs(os.path.join(data_dir, relative_directory_path))
+
+    for relative_file_path in files_to_include:
+        with open(os.path.join(data_dir, relative_file_path), "w"):
+            # Just opening the file is enough to create it.
+            pass
```
jaimemarijke: I'm vaguely remembering some new function that does this more directly

jaimemarijke: ah, Path.touch(), e.g. 

```
from pathlib import Path
Path('/path/to/file.txt').touch()
```

not sure if I like that better

osmo_jupyter/dataset/files_test.py
```diff
@@ -0,0 +1,87 @@
+import os
+import pandas as pd
+
+import osmo_jupyter.dataset.files as module
+
+
+def _init_data_dir(parent_dir, directories_to_include, files_to_include):
+    """set up a data directory with folders and files in it
+    Args:
```
jaimemarijke: nit: missing `parent_dir` here

osmo_jupyter/dataset/files_test.py
```diff
@@ -0,0 +1,87 @@
...
+            tmp_path, "this does NOT exist as a subdirectory"
+        )
+        expected = []
+        assert actual == expected
+
+    def test_gets_file_from_subdirectory_and_ignores_sub_subdirectories(self, tmp_path):
+        _init_data_dir(
+            tmp_path,
+            directories_to_include=["ysi_proodo/subdirectory"],
+            files_to_include=["ysi_proodo/somethin"],
```
jaimemarijke: nit: could add a .txt extension or something to make it clearer that this is a file.

osmo_jupyter/dataset/files.py
jaimemarijke: maybe this could just use a comment with an example folder structure

osmo_jupyter/dataset/files.py
```diff
@@ -0,0 +1,54 @@
+import os
```
jaimemarijke: nit: the name `files.py` seems a little vague. maybe `raw_files` or `get_files`?

jaimemarijke (approved):

rush340 (approved):

osmo_jupyter/dataset/files.py
rush340: I agree that the `FILE_TYPES` name is a little confusing

osmo_jupyter/dataset/files_test.py
jason-curtis: holy mackerel. this is the library I always wanted but never knew I did. I spent altogether too much time in the last few workdays typing out `os.path.join()` calls.
prepare for a changeset that is overly psyched about this library

osmo_jupyter/dataset/files.py
jason-curtis: renamed to source_files

osmo_jupyter/dataset/files.py
jason-curtis: renamed to `FILE_TYPE_SUBFOLDERS` and added some documentation

jason-curtis:
I switched to PosixPaths for everything. upside: that interface is much nicer to use than `os.path` when doing anything file-y. And in most cases it is backwards compatible. I think it's good in this case because we're expecting to do stuff programmatically with these paths almost immediately downstream.

Downside: when you save the output DataFrame as a CSV, the paths are wrapped in `PosixPath(...)` which would complicate deserialization. But you'd already have issues serializing/deserializing the lists that are in here. I think that's OK but we'll have to be aware of it if we ever try to serialize/deserialize this.

osmo_jupyter/dataset/source_files_test.py
```diff
@@ -0,0 +1,86 @@
...
+        assert actual == expected
+
+    def test_gets_file_from_subdirectory_and_ignores_sub_subdirectories(self, tmp_path):
+        _init_data_dir(
+            tmp_path,
+            directories_to_include=["ysi_proodo/subdirectory"],
+            files_to_include=["ysi_proodo/somethin.csv"],
+        )
+        actual = module._get_experiment_data_file_paths_for_type(tmp_path, "ysi_proodo")
+        expected = [tmp_path / "data/ysi_proodo/somethin.csv"]
```
jaimemarijke: nit: weird mix of Path and string

osmo_jupyter/dataset/source_files_test.py
```diff
@@ -0,0 +1,86 @@
...
+            tmp_path, "this does NOT exist as a subdirectory"
+        )
+        expected = []
+        assert actual == expected
+
+    def test_gets_file_from_subdirectory_and_ignores_sub_subdirectories(self, tmp_path):
+        _init_data_dir(
+            tmp_path,
+            directories_to_include=["ysi_proodo/subdirectory"],
+            files_to_include=["ysi_proodo/somethin.csv"],
```
jaimemarijke: also these aren't Paths

jaimemarijke (approved):
I'm okay with the change to using pathlib. not totally convinced by it yet but definitely seems worth trying. my only request would be to go 100% in that direction rather than mixing in string paths

rush340 (approved):
pathlib seems nice!
🚢

EvanSimpson (approved):

PR #39, opened 2019-11-08
Dataset pipeline tools

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1147536099838393)

Adds some tools to help clean up source data files more-so than the existing parse functions. Also adds a function to append experiment metadata to image file names.

I ended up cutting scope on a lot of inconsistencies I found in our historical data, that _should not_ appear going forward.

Things explicitly not supported:
 - kPA barometric pressure
- multiple YSI units for one experiment / in one csv file
- filtering out unequilibrated images
- Multiple source file types for one experiment (calibration + YSI, multiple YSI Formats)
- YSI files that do not conform to the normal column mappings
- Any simultaneous _Experiments_ should be part of separate _Attempts_ (prevent timestamp index collisions)

There's also an explicit dependency that any experiment that shows up in the data collection log needs to have a folder present locally in the local sync directory. I didn't think it was worth working around this since we have ticketed work to replace that with just getting listings directly from S3.

## Manual Testing
I prototyped building a dataset of everything currently on deepthought using these helper functions, and it should give a decent idea of how much work is left to finish this off (plus replacing local image search with S3).

[Notebook link](http://deepthought:8888/notebooks/Experiments/2019-11-08%20Librarify%20More%20Dataset%20Generation%20Process/2019-11-08%20Prototype%20Dataset%20Pipeline%20Tools.ipynb)

For those curious, 31 attempts resulted in about 35k data points and roughly ~9MB (though I think there might be more opportunities to cut out some columns if necessary.


osmo_jupyter/dataset/combine.py
```diff
@@ -204,78 +178,23 @@ def filter_equilibrated_images(equilibration_range: pd.Series, df: pd.DataFrame)
...
-    experiment_names: List[str],
-    calibration_log_filepaths: List[str],
-    picolog_log_filepaths: List[str],
-    process_experiment_result_filepaths: List[str],
-    ROI_names: List[str] = None,
-    msorm_types: List[str] = None,
+def get_images_by_experiment(
+    experiment_metadata: pd.Series, local_sync_directory
 ) -> pd.DataFrame:
+    """ Get all image file names with associated experiment metadata.
```
jason-curtis: this doc doesn't seem right

osmo_jupyter/dataset/source_files.py
```diff
@@ -39,6 +41,12 @@
     "summary_movies",
 ]
 
+VALID_FILETYPES = [".csv", ".mp4", ".gif"]
+
+
+def _check_filepath_is_valid(filepath):
```
jason-curtis: nit: validity doesn't seem quite like the right thing to be checking here, and "data" seems like a helpful thing to mention 
```suggestion
def _is_data_filepath(filepath):
```

osmo_jupyter/dataset/source_files.py
```diff
@@ -72,3 +82,30 @@ def get_experiment_data_files_by_type(project_directory):
             for file_type in FILE_TYPE_SUBFOLDERS
         }
     )
+
+
+def get_all_experiment_images(
```
jason-curtis: note to readers: looks like this was just moved from `.combine`

osmo_jupyter/dataset/parse.py
```diff
@@ -22,8 +22,8 @@ def _apply_parser_configuration(
     )
 
 
-def parse_ysi_kordss_file(filepath: str) -> pd.DataFrame:
-    """ Open and format a YSI KorDSS formatted csv file, with standardized datetime parsing
+def parse_ysi_prosolo_file(filepath: str) -> pd.DataFrame:
```
jaimemarijke, line 92: nice, I'd been thinking we should make this change

osmo_jupyter/dataset/source_files.py
```diff
@@ -39,6 +41,12 @@
     "summary_movies",
 ]
 
+VALID_FILETYPES = [".csv", ".mp4", ".gif"]
```
jaimemarijke: might want to add a comment explaining why we expect each of these. for example, I'm not sure why we'd want anything other than .csvs when we are building the dataset

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
+import datetime
```
jason-curtis: minor: I'm not a huge fan of the name `transform` for this file. It's very generic. It feels a bit more like `load`ing or `normaliz`ing files? don't love those either.

jason-curtis: maybe the problem is that we're trying to do too much with module names. the challenge I was having here is that I saw the filename and thought "well that could be anything". The logical line between this and the `parse` module is fuzzy to me, since that one is also doing some normalization. maybe we should start using module-level docstrings to explain what each module is for?

osmo_jupyter/dataset/source_files.py
```diff
@@ -72,3 +82,30 @@ def get_experiment_data_files_by_type(project_directory):
...
+            for experiment_name in experiment_names
+            for image in os.listdir(os.path.join(local_sync_directory, experiment_name))
+        ],
+        # Ensure correct dtype and column names when no images are found
+        columns=["experiment", "image"],
+        dtype="object",
+    )
+
+    # Filter out experiment log files
+    return all_images.drop(all_images[~all_images["image"].str.contains("jpeg")].index)
```
jaimemarijke: minor: Could this filtering happen in the list comprehension instead?

```
all_images = [
            ...
            for file in os.listdir(os.path.join(local_sync_directory, experiment_name))
            if file.endswith('.jpeg')
        ],
```

osmo_jupyter/dataset/source_files.py
```diff
@@ -72,3 +82,30 @@ def get_experiment_data_files_by_type(project_directory):
...
+
+        Args:
+            local_sync_directory: The local data directory, usually ~/osmo/cosmobot-data-sets
+            experiment_names: A list of experiment directory names in the local sync directory.
+        Returns:
+            DataFrame of all image file names and the corresponding experiment name.
+    """
+    all_images = pd.DataFrame(
+        [
+            {"experiment": experiment_name, "image": image}
```
jaimemarijke: nit: seems like we might want to be consistent about using either `experiment` or `experiment_name`

osmo_jupyter/dataset/source_files.py
jaimemarijke: also, nit: "valid" doesn't seem like quite the right word here, maybe just because it's not clear to me what makes something valid or not. maybe this could be `SOURCE_DATA_FILETYPES`

jason-curtis (approved):

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+import pandas as pd
+
+from .parse import (
+    TIMESTAMP,
+    DO_PCT,
+    DO_MGL,
+    BAROMETRIC_PRESSURE_MMHG,
+    TEMPERATURE_C,
```
jaimemarijke: Maybe these should live in a separate shared file, e.g. `constants.py`. It doesn't seem like `parse` has any more right to own them than `transform` does

jaimemarijke: ohh, also I assumed these were constant numbers but from context here it looks like they are labels? that should probably be in the name somehow, e.g. `DO_PCT_LABEL` or something

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+    return df.drop(columns=columns_to_drop)
+
+
+def _generate_time_based_setpoint_ids(dataset):
+    """Create setpoint IDs that increment whenever there is a > 5 minute gap
+    between image captures."""
+
+    # this assumes the dataset is already sorted by timestamp, so let's assert that
+    pd.testing.assert_frame_equal(dataset, dataset.sort_index())
```
jaimemarijke: since you're going to the work to get the sorted version to compare to, maybe this should just do the sorting and not worry whether it was or wasn't already sorted?

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+def _remove_unused_columns(df):
+    columns_to_drop = [
+        column_name for column_name in df.columns if column_name not in DATASET_COLUMNS
+    ]
+
+    return df.drop(columns=columns_to_drop)
+
+
+def _generate_time_based_setpoint_ids(dataset):
```
jaimemarijke: minor/major: hmm.... this seems a bit brittle, since it relies on our existing practice of always staying at a setpoint for 5 min, which might change. 

Alternate (more complicated) idea: determine setpoint IDs by grouping temporally sequential rows that have "the same" DO/temperature value to within some threshold.

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
...
+        minutes=5
+    )
+    setpoint_ids = setpoint_transitions.astype(int).cumsum()
+    return setpoint_ids
+
+
+def _prepare_ysi_data(ysi_data: pd.DataFrame) -> pd.DataFrame:
+    ysi_data["YSI DO (mmHg)"] = _calculate_partial_pressure(
+        do_percent_saturation=ysi_data["YSI DO (%)"],
+        barometric_pressure_mmhg=ysi_data["YSI barometric pressure (mmHg)"],
```
jaimemarijke: should these use the `DO_PCT` and `BAROMETRIC_PRESSURE_MMHG ` constants?

osmo_jupyter/dataset/transform.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def process_ysi_proodo_file(filepath: str) -> pd.DataFrame:
+    ysi_data = parse_ysi_proodo_file(filepath)
+    return _prepare_ysi_data(ysi_data)
+
+
+def process_ysi_prosolo_file(filepath: str) -> pd.DataFrame:
+    ysi_kordss_data = parse_ysi_prosolo_file(filepath)
+    return _prepare_ysi_data(ysi_kordss_data)
```
jaimemarijke: s/kordss/prosolo

jaimemarijke (approved):

osmo_jupyter/dataset/source_files.py
jason-curtis: (came back here as I noticed this issue in the other PR)
major: checking that the filename contains `jpeg` seems risky since it'll accept files like `not_a_jpeg.txt`. It should be pretty easy to do something a little more airtight, checking that filenames end with `.jpeg` or at least including the dot in the `contains` check.

EDIT: it looks like you can use [`Series.str.contains()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.contains.html) with `regex=True` or even [`Series.str.endswith()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.endswith.html)

jason-curtis (changes requested):
noticed an issue when looking at the other review that I think merits a change

osmo_jupyter/dataset/combine.py
```diff
@@ -204,78 +178,23 @@ def filter_equilibrated_images(equilibration_range: pd.Series, df: pd.DataFrame)
...
+        datetime_from_filename
     )
 
-    equilibrated_data = equilibrated_sensor_data.join(
-        calibration_data.resample("s").interpolate(method="slinear").dropna(axis=1),
-        how="inner",
-    ).sort_index()
-
-    all_images = get_all_experiment_images(local_sync_directory, experiment_names)
+    images_by_experiment["cartridge_id"] = experiment_metadata["cartridge_id"]
```
jaimemarijke: I'm a little confused how this works. I'm assuming the `experiment_metadata` df has one row per experiment, so I'm not clear how this propagates over to the `images_by_experiment` df, which has multiple rows per experiment.

osmo_jupyter/dataset/source_files.py
jaimemarijke: maybe this should be `get_all_experiment_image_filenames` since it doesn't actually get images, and the important thing is that it's finding all the filenames

osmo_jupyter/dataset/combine.py
```diff
@@ -204,78 +178,23 @@ def filter_equilibrated_images(equilibration_range: pd.Series, df: pd.DataFrame)
...
 
-    picolog_data = pd.concat(
-        [
-            parse_picolog_file(picolog_filepath)
-            .resample("s")
-            .interpolate(method="slinear")
-            for picolog_filepath in picolog_log_filepaths
-        ],
-        sort=True,
+    images_by_experiment = get_all_experiment_images(
```
jaimemarijke: (I started making this comment on your other PR, but realized it fit better here).

change requested: some renaming of these functions.

It strikes me that without reading the specifics of the two functions, `get_all_experiment_images ` and `get_images_by_experiment` sound very similar, and it's hard to know what to expect from them. 

I think `get_images_by_experiment` is a bit of a misnomer, since that implies that it returns something grouped or keyed by experiment, which it doesn't. Also, aside from calling `get_all_experiment_images`, the actual logic in this function looks like it has more to do with the other experiment metadata than the images.

osmo_jupyter/dataset/combine.py
jaimemarijke: +1 could definitely use more documentation here.

osmo_jupyter/dataset/combine.py
jaimemarijke: Ah, I believe my confusion stems from the fact that `experiment_metadata` is actually a series - basically just a single row from the data collection log, which might have multiple "experiments" but a single cartridge, cosmobot, etc. 

Maybe it would be more correct to call this `attempt_metadata`, since I believe we've settled on "attempt" as the terminology for a grouping of "experiments"

jaimemarijke (changes requested):

osmo_jupyter/dataset/source_files.py
jaimemarijke: +1 I meant to call out the part about .endswith vs .contains as something to change even if you choose not to move the filtering itself to the list comprehension

osmo_jupyter/dataset/source_files.py
jaimemarijke: +1 I had a similar comment about the module constant name

osmo_jupyter/dataset/transform.py
rush340: It's only relying on equilibration between setpoints taking >5 mins and the time between consecutive image captures at a setpoint being <5 mins. Maybe a comment about those assumptions would help.

It still feels a little brittle. I think I called that out on Slack when I originally went with this method in the notebook. I originally had it doing the alternate idea you're describing, but that won't differentiate the identical setpoints at the point where a sweep reversed.

osmo_jupyter/dataset/transform.py
EvanSimpson: I also struggled a lot with this and really don't know what to do. I don't think this is `load`ing  because that has a specific meaning in data pipelining in "Extract-Transform-Load". "Load" for us would seemingly just be calling `.to_csv` and committing the file. We have "Extract" partially split up between `source_files` and `parse` and then "Transform" split amongst `parse`, `transform`, and `combine`.

It's a real mess.

I think the best option is to just move the `transform` code back into `parse`, since it's not much and they do thematically similar things.

osmo_jupyter/dataset/combine.py
EvanSimpson: I renamed one of these to use `attempt` rather than `experiment` so the distinction should be clearer now.

osmo_jupyter/dataset/transform.py
jason-curtis: yeah I have the same thought about ETL also. I'm not a fan of that acronym though it *is* catchy. Looks like you merged 'em, I'll take a look.

osmo_jupyter/dataset/parse.py
```diff
@@ -175,3 +242,65 @@ def parse_data_collection_log(filepath: str) -> pd.DataFrame:
...
+
+    ysi_data_resampled = ysi_data.resample("s").interpolate(method="slinear")
+    ysi_data_rounded = ysi_data_resampled.round(6)
+
+    return _remove_unused_columns(ysi_data_rounded)
+
+
+def process_ysi_proodo_file(filepath: str) -> pd.DataFrame:
+    """ Parse a YSI ProODO data csv as a DataFrame with only the columns necessary for deep learning.
+        Numerical values are rounded.
```
jason-curtis, line 261: :+1: I like this framing of these functions. It feels about right for now that the functions that do the most are `process_*_file()`

osmo_jupyter/dataset/parse_test.py
```diff
@@ -168,3 +169,84 @@ def test_get_attempt_summary_gets_multiple_buckets(self):
...
+                # Interpolated value
+                pd.to_datetime("2019-01-01 00:00:01"),
+                pd.to_datetime("2019-01-01 00:00:02"),
+            ],
+            "YSI DO (%)": [50, 50.5, 51],
+            "YSI DO (mg/L)": [3, 3.000015, 3.00003],
+            "YSI temperature (C)": [29, 29.5, 30],
+            "YSI barometric pressure (mmHg)": [750, 752.5, 755],
+            # New column with derived-then-interpolated partial pressure values
+            # Rounded to at most 6 sigfigs
```
jason-curtis, line 249: minor: I don't think [`pd.testing.assert_frame_equal`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html) checks this deep by default. I think it only checks the first 5 decimal points unless you ask for something different.

jason-curtis (approved):

rush340 (approved):

PR #41, opened 2019-11-12
Fix dataset helper function issues and use S3 to get experiment images

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1148349932180745)

This is a diff on #39 - we could either review them together, or go through the review process for #39 first, then tackle this one after it's been merged. I don't have a strong preference.

Notably, I removed the AWS credentials from the call to `boto.connect_s3()` so there's now a strict dependency that there be credentials in the calling env / a `.aws/credentials` file. This shouldn't be a huge issue since both workstation have that, and this will most likely be called from workstation jupyter notebooks.

### Bugfixes
* Fixes intermediate image-experiment dataframe column name usage.
* Interpolates calibration data to fill in gaps in sensor readings to align to more images.
* Stop adding setpoint IDs to raw calibration data - it's meant to be used on image timestamps.

## Manual Testing
Tried my prototype notebook locally with this change - wound up with about 38k images in total.


osmo_jupyter/dataset/source_files.py
```diff
@@ -1,9 +1,12 @@
-import os
 from pathlib import Path
 from typing import List
 
+import boto
 import pandas as pd
 
+CAMERA_SENSOR_EXPERIMENTS_BUCKET_NAME = "camera-sensor-experiments"
```
jaimemarijke: :P "camera sensor experiments" is a kind of old name. maybe the constant should just be `EXPERIMENTS_BUCKET_NAME`

osmo_jupyter/dataset/source_files.py
```diff
@@ -109,3 +110,38 @@ def get_all_experiment_images(
...
+    s3_prefix = f"{experiment_directory}/"
+    all_keys = list_camera_sensor_experiments_s3_bucket_contents(s3_prefix)
+    prefix_length = len(s3_prefix)
+    filenames = [key[prefix_length:] for key in all_keys]
+    return filenames
+
+
+# COPY PASTA - modified from cosmobot-process-experiment
+# removed S3 credentials - must be present in the local env to work
+def list_camera_sensor_experiments_s3_bucket_contents(
```
jaimemarijke: nit: similar comment as above - I'm thinking we should phase out the "camera sensor experiments" name, since that's not really how we talk about it anymore, and just use "experiments" or "cosmobot experiments" instead

update: I didn't realize this was the original fn name. I suppose there's something to be said to try to keep the copy-pasta'd functions as similar as possible to facilitate future refactoring, but right now I'm thinking it would be better to just factor and name these functions however we want them to be for this code.

jaimemarijke: other nit: could be underscore-prefixed

osmo_jupyter/dataset/source_files.py
```diff
@@ -109,3 +110,38 @@ def get_all_experiment_images(
 
     # Filter out experiment log files
     return all_images.drop(all_images[~all_images["image"].str.contains("jpeg")].index)
+
+
+# COPY PASTA - from cosmobot-process-experiment
```
jason-curtis: it'd be nice to be more specific here and below
```suggestion
# COPY PASTA - from cosmobot-process-experiment - osmo_camera.s3
```
could even include the revision # in there...

osmo_jupyter/dataset/source_files.py
```diff
@@ -109,3 +110,38 @@ def get_all_experiment_images(
...
+    Returns:
+        list of key names under the prefix provided.
+    """
+    try:
+        s3 = boto.connect_s3()
+    except boto.exception.NoAuthHandlerFound:  # type: ignore
+        print(
+            "You must have aws credentials already saved, e.g. via `aws configure`. \n"
+        )
+        return []
```
jason-curtis: major: This seems like a critical enough error that it should explode, no?

jason-curtis (approved):
nice work removing the creds. I don't remember why those were living there but I think must have been having some initial setup issue.

jason-curtis (changes requested):
previous review should have been "request changes" because of the one error handling thing

jaimemarijke (approved):

osmo_jupyter/dataset/source_files.py
jason-curtis: hmm I'm thinking this now too. there isn't really that much copypasta'd code and much some of it is modified. *maybe* (just maybe) we should just not be apologetic about having similar, but different code in two different packages for two different purposes.

rush340 (approved):

osmo_jupyter/dataset/source_files.py
rush340: What if it were [three](https://github.com/OsmoSystems/cosmobot-run-experiment/blob/master/cosmobot_run_experiment/s3.py#L46) different packages?  :P 

I'm all for a better function name. We could put the original in another comment above it if we want traceability.

EvanSimpson:
I added in a bugfix+test fix for an issue I found in the last PR.

osmo_jupyter/dataset/combine.py
```diff
@@ -192,7 +190,6 @@ def get_all_attempt_image_filenames(
                 * cartridge_id
                 * cosmobot_id
                 * pond
-            local_sync_directory: The local sync directory in which to look for image files.
         Returns:
             A DataFrame of images filenames along with timestamp, experiment name, and
```
jason-curtis: since we're having issues with column names, maybe this docstring should be more specific about column names?

osmo_jupyter/dataset/source_files.py
```diff
@@ -84,14 +87,11 @@ def get_experiment_data_files_by_type(project_directory):
...
-) -> pd.DataFrame:
+def get_all_experiment_image_filenames(experiment_names: List[str]) -> pd.DataFrame:
     """
         Get a DataFrame of all image files across multiple experiment data directories.
 
         Args:
-            local_sync_directory: The local data directory, usually ~/osmo/cosmobot-data-sets
             experiment_names: A list of experiment directory names in the local sync directory.
         Returns:
             DataFrame of all image file names and the corresponding experiment name.
```
jason-curtis: same note about this docstring - could be more specific on column names

jaimemarijke (approved):

osmo_jupyter/dataset/parse.py
```diff
@@ -62,16 +62,16 @@ def _remove_unused_columns(df):
...
 
         This is dependent upon equilibration between setpoints taking >5 mins and the time between
         consecutive image captures at a setpoint being <5 mins.
     """
 
-    calibration_log_sorted = calibration_log_data.sort_index()
+    image_data_sorted = image_data.sort_index()
 
-    setpoint_transitions = calibration_log_sorted.index.to_series().diff() > datetime.timedelta(
+    setpoint_transitions = image_data_sorted.index.to_series().diff() > datetime.timedelta(
```
jason-curtis: previously it was leaning on a concept of a `calibration_log` convention which kiiiinda implied indexing by timestamp, but with the name change (`calibration_log_data` -> `image_data`) it's even less clear that we expect this to be indexed on timestamp.

Maybe we should just change this function to take a Series of timestamps?

osmo_jupyter/dataset/parse.py
```diff
@@ -295,12 +295,12 @@ def process_calibration_log_file(filepath: str) -> pd.DataFrame:
...
 
-    rounded_calibration_data = trimmed_calibration_data.round(
-        {"setpoint temperature (C)": 3, "setpoint O2 fraction": 6}
+    calibration_data_resampled = trimmed_calibration_data.resample("s").interpolate(
+        method="slinear"
     )
 
-    rounded_calibration_data["setpoint ID"] = _generate_time_based_setpoint_ids(
-        rounded_calibration_data
+    rounded_calibration_data = calibration_data_resampled.round(
```
jason-curtis, line 308: did you mean to remove setpoint ID generation here? if so we should remove it from the docstring too.

osmo_jupyter/dataset/parse_test.py
```diff
@@ -171,20 +171,17 @@ def test_get_attempt_summary_gets_multiple_buckets(self):
         pd.testing.assert_series_equal(actual_attempt_summary, expected_attempt_summary)
 
 
-def test_processes_log_data_correctly(mocker):
+def test_intertoplates_calibration_log_data_correctly(mocker):
```
jason-curtis:
```suggestion
def test_interpolates_calibration_log_data_correctly(mocker):
```

osmo_jupyter/dataset/combine_test.py
```diff
@@ -297,12 +297,16 @@ class TestGetImagesByExperiment:
     def test_combines_experiment_metadata_correctly(self, mocker):
         mock_image_data = pd.DataFrame(
             {
-                "experiment": [
+                "experiment_name": [
```
jason-curtis, line 300: is there anything downstream (jupyter snippets, maybe) that needs to know about these column name changes?

jason-curtis (changes requested):

osmo_jupyter/dataset/parse.py
EvanSimpson: As per offline discussion - yes setpoint ID generation as implemented above needs to be applied to image timestamps, and not calibration timestamps.

osmo_jupyter/dataset/combine_test.py
EvanSimpson: I don't think so, so far. Would only affect the end consumer, of which there aren't any yet.

jason-curtis (approved):

rush340 (approved):

PR #42, opened 2019-11-19
Bugfix: Pivot process experiment results before concat

EvanSimpson:
[Asana Ticket](https://app.asana.com/0/819671808102776/1148728220755426)

If we concat multiple experiment results files before pivoting, the pivot blows up if there are duplicate timestamps in the index (as there are when we collect on multiple units simultaneously). 

Fix here is to pivot each file individually and then concat them.

## Manual Testing

Ran the modified function with a list of result files from attempt 19 (units E and D collected data simultaneously). Checked that the output dataframe had the same number of rows as if each file had been passed in individually and concatenated afterward, also verified that the combined dataframe had rows for distinct images taken during the same second.

Screen Shot 2019-11-19 at 11 21 30 AM

### DIFF
--- a/osmo_jupyter/dataset/combine.py
+++ b/osmo_jupyter/dataset/combine.py
@@ -86,7 +86,7 @@ def get_equilibration_boundaries(equilibration_status: pd.Series) -> pd.DataFram
 def pivot_process_experiment_results_on_ROI(
     experiment_df: pd.DataFrame,
     ROI_names: List[str] = None,
-    msorm_types: List[str] = None,
+    pivot_values: List[str] = None,
 ) -> pd.DataFrame:
     """
         Flatten a DataFrame of process experiment results down to one row per image.
@@ -94,13 +94,14 @@ def pivot_process_experiment_results_on_ROI(
         Args:
             experiment_df: A DataFrame of process experiment summary statistics.
             ROI_names: Optional. A list of ROI names to select. Defaults to all ROIs present.
-            msorm_types: Optional. A list of MSORM column names to select. Defaults to all RGB channel MSORMs.
+            pivot_values: Optional. A list of column names to select for each ROI in the pivot.
+                Defaults to all RGB channel MSORMs.
         Returns:
             DataFrame with one row per image, and msorm values for a subset of ROIs. Creates one column
             for every unique ROI in the dataset. NaN values are used when an image is missing an ROI.
     """
-    if msorm_types is None:
-        msorm_types = ["r_msorm", "g_msorm", "b_msorm"]
+    if pivot_values is None:
+        pivot_values = ["r_msorm", "g_msorm", "b_msorm"]
 
     if ROI_names:
         experiment_df = experiment_df[experiment_df["ROI"].isin(ROI_names)]
@@ -112,7 +113,7 @@ def pivot_process_experiment_results_on_ROI(
 
     # Pivot creates a heirachical multiindex data frame with an index for each 'values' column
     pivot = experiment_df.pivot(
-        index="timestamp", columns="ROI", values=msorm_types + ["image"]
+        index="timestamp", columns="ROI", values=pivot_values + ["image"]
     )
 
     # After pivot -> Multi-level index
@@ -141,7 +142,7 @@ def pivot_process_experiment_results_on_ROI(
 def open_and_combine_process_experiment_results(
     process_experiment_result_filepaths: List[str],
     ROI_names: List[str] = None,
-    msorm_types: List[str] = None,
+    pivot_values: List[str] = None,
 ) -> pd.DataFrame:
     """
         Open multiple process experiment result files and combine into a single DataFrame with one row per image.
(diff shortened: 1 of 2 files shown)


osmo_jupyter/dataset/combine_test.py
```diff
@@ -266,6 +266,25 @@ def test_combines_image_rows_by_ROI(self):
...
+
+        # Open the same file twice to ensure there are duplicate timestamps
+        pivot_results = module.open_and_combine_process_experiment_results(
+            process_experiment_result_filepaths=[
+                test_process_experiment_file_path,
+                test_process_experiment_file_path,
+            ]
+        )
+
+        # Bonus assertion - make sure we preserve duplicate rows for each timestamp
```
jason-curtis: rather than having a 'bonus' it might be clearer to just be more specific about the behavior you want in the title!

osmo_jupyter/dataset/combine.py
```diff
@@ -141,29 +142,34 @@ def pivot_process_experiment_results_on_ROI(
 def open_and_combine_process_experiment_results(
     process_experiment_result_filepaths: List[str],
     ROI_names: List[str] = None,
-    msorm_types: List[str] = None,
+    pivot_values: List[str] = None,
```
jason-curtis: `pivot_columns` or `pivot_column_names` would be a little clearer to me here.

osmo_jupyter/dataset/combine.py
```diff
@@ -86,21 +86,22 @@ def get_equilibration_boundaries(equilibration_status: pd.Series) -> pd.DataFram
...
         Flatten a DataFrame of process experiment results down to one row per image.
 
         Args:
             experiment_df: A DataFrame of process experiment summary statistics.
             ROI_names: Optional. A list of ROI names to select. Defaults to all ROIs present.
-            msorm_types: Optional. A list of MSORM column names to select. Defaults to all RGB channel MSORMs.
+            pivot_values: Optional. A list of column names to select for each ROI in the pivot.
+                Defaults to all RGB channel MSORMs.
         Returns:
             DataFrame with one row per image, and msorm values for a subset of ROIs. Creates one column
```
jason-curtis: this is no longer just MSORMs.`msorm values` -> `pivot-column values`

jason-curtis (approved):

osmo_jupyter/dataset/combine.py
```diff
@@ -86,21 +86,22 @@ def get_equilibration_boundaries(equilibration_status: pd.Series) -> pd.DataFram
...
             for every unique ROI in the dataset. NaN values are used when an image is missing an ROI.
     """
-    if msorm_types is None:
-        msorm_types = ["r_msorm", "g_msorm", "b_msorm"]
+    if pivot_values is None:
+        pivot_values = ["r_msorm", "g_msorm", "b_msorm"]
```
jaimemarijke: seems like this could just be the default value of the parameter. could use a tuple so it's not mutable

osmo_jupyter/dataset/combine.py
jaimemarijke: +1

osmo_jupyter/dataset/combine.py
```diff
@@ -141,29 +142,34 @@ def pivot_process_experiment_results_on_ROI(
...
-            msorm_types: Optional. A list of MSORM column names to select. Defaults to all RGB channel MSORMs.
+            pivot_values: Optional. A list of column names to select for each ROI in the pivot.
+                Defaults to all RGB channel MSORMs.
         Returns:
             DataFrame of all summary statistics flattened to one row per image with all selected ROIs.
     """
-    if msorm_types is None:
-        msorm_types = ["r_msorm", "g_msorm", "b_msorm"]
+    if pivot_values is None:
+        pivot_values = ["r_msorm", "g_msorm", "b_msorm"]
```
jaimemarijke: maybe DRY up with lines 102-103? could this default be defined in only one function?
(also I see now this is not really your change, since you just changed the name)

jaimemarijke (approved):

osmo_jupyter/dataset/combine.py
EvanSimpson: I went with removing the default from the outer function, but left the list default as-is in the inner, so that it still has the chance to override the default None passed in from the outer function. Seemed like the better tradeoff to having the tuple default in both.

osmo_jupyter/dataset/combine.py
EvanSimpson: I guess I went a little too literal here - it's passed into the `pivot` function as the `values` arg.

osmo_jupyter/dataset/combine_test.py
EvanSimpson: But that's slightly less whimsical!

osmo_jupyter/dataset/combine.py
EvanSimpson: See comment above.

rush340 (approved):

PR #44, opened 2019-11-21
Parse data collection log with missing s3 buckets

EvanSimpson:
Related to [Automate adding attempts into dataset repo](https://app.asana.com/0/819671808102776/1148349932180749)

In putting together the notebook for the above ticket I ran into an issue with parsing the data collection log in its current state, where some rows existed for in-progress data collection with an unpopulated S3 Bucket(s) column. I figured this would be a recurring problem so I decided to take care of it in code rather than try to come up with a useful placeholder.

## Manual Testing
Used it against a snapshot of the current data collection log which exhibits this issue.

### DIFF
--- a/osmo_jupyter/dataset/parse.py
+++ b/osmo_jupyter/dataset/parse.py
@@ -198,8 +198,13 @@ def parse_calibration_log_file(filepath: str) -> pd.DataFrame:
 
 
 def _get_attempt_summary(attempt: pd.Series) -> pd.Series:
-    experiment_names = attempt["S3 Bucket(s)"].split("\n")
+    if attempt.notna()["S3 Bucket(s)"]:
+        experiment_names = attempt["S3 Bucket(s)"].split("\n")
+    else:
+        experiment_names = []
+
     pond = attempt["Scum Tank"] if "Scum Tank" in attempt.index else "calibration"
+
     return pd.Series(
         {
             "experiment_names": experiment_names,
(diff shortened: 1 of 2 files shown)


jason-curtis (approved):

rush340 (approved):

osmo_jupyter/dataset/parse.py
```diff
@@ -198,8 +198,13 @@ def parse_calibration_log_file(filepath: str) -> pd.DataFrame:
 
 
 def _get_attempt_summary(attempt: pd.Series) -> pd.Series:
-    experiment_names = attempt["S3 Bucket(s)"].split("\n")
+    if attempt.notna()["S3 Bucket(s)"]:
```
jaimemarijke: nit: `if attempt["S3 Bucket(s)"] is not None` seems clearer to me

I also have a slight personal preference for the ternary syntax because the variable only has to be defined once, e.g.

```
experiment_names = (
    attempt["S3 Bucket(s)"].split("\n")
    if attempt["S3 Bucket(s)"] is not None
    else []
)
```

jaimemarijke (approved):

osmo_jupyter/dataset/parse.py
EvanSimpson: It looks like the pandas default nan/null value `np.nan` rather than `None`, but otherwise this is much clearer!

EvanSimpson: I ended up going with `pd.isnull()` which will check for all sorts of null values.

